PR #142, opened 2015-01-01
:bind-keymap - bind a key prefix to an autoloaded package keymap

killdash9:
John, 

Please consider this patch.  I've added the ability to bind a key sequence to a package's keymap.  Package loading is deferred until that key sequence has been pressed, at which time the package is transparently loaded so that the next key pressed is routed into the package's keymap.  

Here's an example of its use:
```
(use-package ahg :bind-keymap ("C-c h g" . ahg-global-map))
```

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -189,6 +189,8 @@ ARCHIVE can be a string or a symbol or 'manual to indicate a manually updated pa
     :pre-init
     :pre-load
     :requires
+    :bind-keymap
+    :bind-keymap*
     )
   "Keywords recognized by `use-package'.")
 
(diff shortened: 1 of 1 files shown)


jwiegley:
@killdash9 It looks alright, I just wonder why I never came across a need for something like this.  Do others have any comments?

killdash9:
For packages with just a few commands, it's easy enough to specify individual :bind's for them.  :bind-keymap is useful for packages with a larger number of commands placed behind a common prefix.  js2r (javascript refactoring library) is another one that comes to mind.

PR #145, opened 2015-01-03
Passing t into keymap function

killdash9:


killdash9:
This was an oversight

Issue #149, opened 2015-01-12
Introduced runtime dependency: use-package-verbose


PhilHudson:
See issues #13 and #17 -- it's happened again, culprit this time is `use-package-verbose`. Effect is that you have to `(require use-package)` at runtime or you get a void-variable error. As before, the fix will be to ensure the variable (user option) is eval'ed at code-planting time rather than at runtime.

npostavs:
Introduced by #96.

jwiegley:
@tarsius Can you address this?

tarsius:
> Effect is that you have to (require use-package) at runtime or you get a void-variable error.

That's seems reasonable to me. I don't really understand why it's desirable to use `use-package` without requiring it at runtime - can you show me the code?

PhilHudson:
Jonas Bernoulli <notifications@github.com> writes:

>     Effect is that you have to (require use-package) at runtime or you get a
>     void-variable error.
>
> That's seems reasonable to me. I don't really understand why it's desirable to
> use use-package without requiring it at runtime - can you show me the code?

Isn't that the whole point of implementing use-package as a compiler
macro: that you *don't* have to load it at runtime?

Not saying that's *got* to be the way it works, but if it is not a major
design goal, then there are downsides to the macro thing that might mean
it was better to use "ordinary" functions instead. I believe that's one
of the standard "prefer x over y" things of Lisp programming: don't use
a macro where a function will do.

-- 
Phil Hudson                   http://hudson-it.ddns.net
@UWascalWabbit                 PGP/GnuPG ID: 0x887DCA63

tarsius:
Again:

> can you show me the code?

I suppose something like this

```
;;; init.el -- my init file

(add-to-list 'load-path "/path/to/use-package")
(eval-when-compile (require 'use-package))

(use-package foo :init (do stuff))
```

?

tarsius:
> Isn't that the whole point of implementing use-package as a compiler macro: that you *don't* have to load it at runtime?

Well I suppose you could look at it that way, then again isn't the whole point of having a variable that changing its value actually has an effect?

tarsius:
> Isn't that the whole point of implementing use-package as a compiler macro: that you *don't* have to load it at runtime?

So? That doesn't mean that the expansion cannot rely on global variables. You just have to load the library - that doesn't result in the macro not being expanded until runtime.

PhilHudson:
Jonas Bernoulli <notifications@github.com> writes:

> You
[...]

PR #151, opened 2015-01-18
Allow using expanded macro without loading feature

tarsius:
In the macro `use-package-with-elapased-timer` use `bound-and-true-p`
go get the values of the customizable options `use-package-verbose`
and `use-package-minimum-reported-time`.  This way the library only
has to be required at compile time, provided these options are not
actually customized.  If the user has changed the values, then she
also has to load the library at runtime or the macros fall back to
the default of doing their job silently.  See #149.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -48,12 +48,27 @@
   :group 'startup)
 
 (defcustom use-package-verbose nil
-  "Whether to report about loading and configuration details."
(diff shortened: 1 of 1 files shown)


npostavs:
Would putting `;;;###autoload` on the `defcustom`s get the same effect?

PR #159, opened 2015-03-10
Eval the argument to :disable

brontitall:
This allows disabling a package conditionally, e.g.:

    (use-package smex
      :disable (< emacs-major-version 24)
      :ensure t
      ...)

or

    (use-package paradox
      :disable (version< emacs-version "24.4")
      :ensure t
      ...)

It might be nice to have a feature of use-package that gracefully dealt with packages that can't be installed, but this allows fairly general logic to be used when disabling a package.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -358,7 +358,7 @@ For full documentation. please see commentary.
          (name-symbol (if (stringp name) (intern name) name)))
 
     ;; force this immediately -- one off cost
-    (unless (use-package-plist-get args :disabled)
+    (unless (eval (use-package-plist-get args :disabled))
 
       (when archive-name
         (use-package-pin-package name archive-name))


npostavs:
What about `:if`?

brontitall:
Yes, I considered `:if`, but it controls only the loading of the package, not the action of `:ensure`. My use case is to have the same config on different versions of Emacs. On Emacs 23 even installing some packages (such as the ones in my examples) fails, due to version dependencies.

Further, `:ensure` is evaluated at compile-time, so even wrapping the `use-package` form in a conditional won't help if you have a byte-compiled config. You end up having to use `eval-when`, as well as the conditional and by that point you have a fer less elegant solution than just code for `:disable`

YoungFrog:
If there's interest for this feature, I suggest to set the
EVAL-BACKQUOTE argument of ```use-package-plist-get``` to ```t```
instead of forcing evaluation. This allows to indicate the need for
evaluation by a comma :
```emacs-lisp
(use-package foo :disabled (eq t t))  ; disabled
(use-package foo :disabled (eq t nil)) ; disabled
(use-package foo :disabled ,(eq t t)) ; disabled
(use-package foo :disabled) ; disabled
(use-package foo :disabled ,(eq t nil)) ; not disabled.
```

jwiegley:
Hmm... that seems a bit subtle and tricky.

brontitall:
This:

> (use-package foo :disabled (eq t nil)) ; disabled

Seems to violate the [POLA](http://en.wikipedia.org/wiki/Principle_of_least_astonishment).

I can't think of a use case where this behaviour would be of use.

YoungFrog:
Le 11/03/2015 10:00, brontitall a écrit :
> This:
> 
>     (use-package foo :disabled (eq t nil)) ; disabled
> 
> Seems to violate the POLA

Having some arguments evaluated while others are not also violates this principle, IMHO. I prefer an explicit request for evaluation (with a comma). I'm not gonna fight for it though.

brontitall:
I understand this position, and I can see its merits in the cause of consistency between the arguments of `use-package`. The arguments to `:init` and `:config` are not evaluated by default at macro expansion time, so why should this one be?

I would only suggest that due to the way it acts the only possible time when the argument to `:disabled` is of any use is at expansion time, so evaluating at that time would have the most utility. In other words, I feel it's more in the style of `:pre-load` than of `:command`.

By the way, I've not seen this behaviour before (where a macro or anything else uses a comma not lexically inside a back-quoted form). In fact, until your comment I was unaware of this behaviour of `:command`, `:diminish`, `:defines`, etc. It doesn't seem to appear in the documentation. Clearly I need to spend more time looking at the code. Maybe I'm missing some idiom that I've not noticed which makes this expected. Is this widely used?

jwiegley:
I also dislike this special comma appearing "out of nowhere", so to speak.  Also `use-package` is a macro, it isn't an explicit back-quoted form, so expecting anyone to realize that they can use commas in most places (though, not all!) is both confusing and a bit too magical.

I would prefer we had complete consistency across all options, and avoided from exposing the internal structure of `use-package`.  That is, `use-package` defines a flexible property list (in that, not every property needs an associated value), where the values assigned to properties are evaluated under different conditions, depending on the nature of that property.

Every property, in my opinion, should allow for a lambda to be specified which indicates that it is to be executed when that property requires a value.  This makes the dynamism of the property's value explicit, and gives us more flexibility in the future.

Most properties should accept multiple flavors of values: a symbol, a string, a list of strings, and a lambda.  It is up to the property to decide what is flavor means.

I would like to see a general mechanism in the code for property evaluation which encodes this policy, as it would then be dead simple to document, and intuitive for new users.

npostavs:
I originally introduced the "magic comma" just because I thought it was less bad than magically evaluating symbols in some positions.

> Every property, in my opinion, should allow for a lambda to be specified which indicates that it is to be executed when that property requires a value.

I think a magic lambda is slightly worse than a magic comma just because the comma is shorter and therefore introduces less clutter, but it doesn't make much difference to me.

jwiegley:
Why would a "magic lambda"?  Each property accepts a single value, and it's up to that property to determine what the value means, and when it is evaluated.  If it's a function, then the evaluation is able to programmatically determine the value.

So, since lambda's are just values, they are then no more magical than any other value a property may accept.  The comma, however, is certainly magical, since that's not a value but an anti-quoted form that is normally only accepted within backquoted forms.

brontitall:
So, @jwiegley am I understanding correctly that for my use case, I would use:

    (use-package foo
      :disabled (lambda () (< emacs-major-version 24))
      ...)

This seems ok, but I agree, it's noisier.

jwiegley:
It's intended that you write:

``` lisp
(use-package foo
  :if (>= emacs-major-version 24)
  ...)
```

We should have `:when` and `:unless` keywords too.

brontitall:
>
> It's intended that you write:
>
> (use-package foo
>   :if (>= emacs-major-version 24)
>   ...)
>
> We should have :when and :unless keywords too.
>
 Ok, but as I mentioned above, :if (currently) doesn't control :ensure

npostavs:
> Since lambda's are just values, they are no more magical than any other values a property might accept. The comma, however, is certainly magical, since it's not a value (per se)

Treating lists that start with `lambda` specially is equally magical to treating lists that start with `,` specially. 

On the other hand, now that I've rechecked, the Elisp manual doesn't explicitly document the way read syntax for unmatched commas works, so I guess that's enough to push me in favour of using `lambda`.

jwiegley:
@brontitall `:disabled` has a very specific, hard-core meaning of: "If this is present in any way, do nothing else at all and ignore this package!"  For logic-based exclusion, please use `:if` or `:unless`.

brontitall:
I acknowledge that my proposed change may not the way to fix this problem, but either I'm missing something or the use-case I describe can not be addressed with `:if` or `:unless`.

These two options don't control `:ensure`. This means that if I try to `use-package` with `:ensure` on a package that doesn't work on e.g. emacs 24.3 (by depending on the virtual package "emacs-24.4" then the `package-install` will still be run, and emacs initialization will fail. Without some conditional that covers `:ensure` there seems to be no neat way to `use-package` in this case.

I see that the documentation has been updated to include:

    :disabled The package is ignored completely, the same as `:if nil'.

This is not accurate. `:disabled` controls whether a package is installed via `:ensure`, `:if` does not.

jwiegley:
@brontitall Thanks for that clarification.

Also, I need to understand what `:ensure` is doing better, to find the most consistent behavior here for `:disable` and `:if`.  I'd like to handle your use case, which may perhaps mean allowing `:ensure` to accept a boolean value, rather than only testing for its presence.

PR #163, opened 2015-03-14
Fix :ensure value interpretation

### LINKED ISSUE #164: :ensure no longer works for me
This code:

```
(use-package ag :ensure t)
```

No longer works for me. Neither does:

```
(use-package ag :ensure ag)
```

Am I using ensure wrong?

This is using the most recent version from MELPA (20150314.322)

thomasf:
Fixes #164

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -234,11 +234,21 @@ then the expanded macros do their job silently."
            (use-package-as-one (symbol-name head) args
              #'use-package-normalize-symbols))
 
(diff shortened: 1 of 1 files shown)


jwiegley:
@thomasf Thanks for the fix, the only issue is that the check for `null args` has to be performed before the call to `use-package-only-one`, since the latter will raise an error if `args` happens to be null.  I'll make the fix here.

PR #165, opened 2015-03-15
Update README.md, :pin now need a string

bricewge:
Since 2.0 `:pin` need the archive name as a string.


jwiegley:
@bricewge Actually, I'd prefer to relax the normalization logic to coerce a symbol passed to `:pin` into a string.

PR #168, opened 2015-03-15
use--package: eval-after-load name

### LINKED ISSUE #167: `:config` seems to be called too early
Hello,

I have the following config:

``` elisp
(use-package fic-mode
  :defer t
  :config
  (setq fic-background-color "Orange")
  (setq fic-foreground-color "Black")
  (modify-face 'font-lock-fic-face fic-foreground-color
               fic-background-color nil t nil nil nil nil))
```

But this seems to execute the `:config` block all the time, like if it was `:init` (the error is `Error (use-package): Failure in :config of fic-mode: (error "Invalid face" font-lock-fic-face)`).

At this point fic-mode is not loaded so the `:config` block should not execute (e.g `(featurep 'fic-mode)` returns nil).

When I `load-library` fic-mode or `M-x fic-mode` the error goes away when evaling the `modify-face` form.

npostavs:
instead of name-string. This prevents triggering by config file names
that have the same name as the package.

Reprise of #53.
Fixes #167.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -336,7 +336,7 @@ the user specified.")
           (ignore
            (display-warning 'use-package (format ,fmt ,err) :error)))))))
 
-(defun use--package (name-symbol name-string args)
+(defun use--package (name name-symbol name-string args)
   "See docstring for `use-package'."
   (let*
       ((commands (plist-get args :commands))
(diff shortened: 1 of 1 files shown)


jwiegley:
@npostavs What if we just always used `name-symbol`?

npostavs:
> What if we just always used name-symbol?

Then it couldn't work for packages that don't use `provide`, but maybe that's okay.

PR #176, opened 2015-03-17
minor fixes to get bind-keymap working in 2.0

killdash9:


use-package.el
```diff
@@ -384,14 +384,14 @@ possible."
                      `(bind-key ,(car binding)
                                 #'(lambda () (interactive)
                                     (use-package-autoload-keymap
-                                     ',(cdr binding) ,name-symbol nil))))
+                                     ',(cdr binding) (quote ,name-symbol) nil))))
```
npostavs: This should probably use `',name-symbol`, for consistency.

PR #199, opened 2015-04-12
bind-keymap fixes.

killdash9:
Override argument no longer ignored.  Not adding keymap to list of
commands. `eval' no longer appears to be necessary, using direct
invocation of bind-key*.


jwiegley:
If you aren't using `:commands`, then deferral will not happen.

killdash9:
`:bind-keymap` can't use `autoload` like `:bind`.  This is because it binds a key to a package's keymap, not to a command.  Instead of `autoload`, `bind-keymap` is implemented by an autoload-like function called `use-package-autoload-keymap`.  I wrote the bind-keymap functionality a few months ago.  It was broken during the refactoring, and this patch fixes it.

PR #200, opened 2015-04-12
Add :el-get keyword similar to :ensure

jonEbird:
Use the :el-get keyword to have a package installed via el-get

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -108,6 +108,7 @@ the user specified."
   '(:disabled
     :pin
     :ensure
+    :el-get
     :if
     :when
     :unless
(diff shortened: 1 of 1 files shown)


jonEbird:
While my patch gets el-get installations working, I'm wondering if you'd prefer to just continue to use `:ensure` with a `:pin el-get` instead?

jwiegley:
I'd much rather that this was placed in its own module, and submitted to `el-get`, rather than having a dependency on el-get within use-package.  There is no way I will ever be able to test this properly.

jonEbird:
That's fine. Could you elaborate on how you might approach putting this into a separate module and submitted to `el-get`? Would it be just updating `use-package-keywords` and defining the necessary functions? Thanks.

jwiegley:
Yes, that should be all that is needed to introduce a new `use-package` keyword from an external package.

edvorg:
Hi. Why do you need explicit :ensure (:el-get) keywords? I mean use package might check for package presence by itself and install it if needed.

thomasf:
@edvorg 
1. specifying package installl names which might differ from their load names
2. selectively overriding built in packages like org-mode
3. not causing additional load/compile times for packages where it's not needed.

edvorg:
Thank you, Thomas.

Issue #189, opened 2015-03-22
Delight addon inclusion


darkfeline:
Delight addon for use-package, request for inclusion (in place of diminish?)

https://gist.github.com/darkfeline/f615f8ccb4cc104ca6a6

This is working for me so far.

Old issue:
https://github.com/jwiegley/use-package/issues/187

jwiegley:
This is great, thank you!  I'm happy have it in addition to diminish, but I don't want to remove that support, since it works just fine for me -- and others have their configurations already based on it.

jwiegley:
@darkfeline Isn't it impossible for the first test in your `cond` form to ever return `t`?

darkfeline:
@jwiegley Yeah, sorry for the bug, should be `(symbolp (car args))`

https://gist.github.com/darkfeline/f615f8ccb4cc104ca6a6

PR #206, opened 2015-04-29
Enhance :delight with possible values string or t

andersjohansson:
Enhances #189.

We now should have

    (use-package my-package
     :delight "rep")

to mean `(delight 'my-package-mode "rep" 'my-package)`, and

     :delight t

to mean `(delight 'my-package-mode nil 'my-package)`. I copied some of the logic from :diminish.

Testing this quickly, I didn't stumble on any problems but it would be great if both @jwigley and @darkfeline could have a look so I didn't miss out anything important from your first implementation.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -933,9 +933,14 @@ deferred until the prefix key sequence is pressed."
 (defun use-package-normalize/:delight (name-symbol keyword args)
   "Normalize arguments to delight."
   (cond
-   ((and (= (length args) 1)
-         (symbolp (car args)))
-    (list (car args) nil name-symbol))
+   ((= (length args) 1)
+    (cond
+     ((eq (car args) t)
+      (list (intern (concat (symbol-name name-symbol) "-mode")) nil name-symbol))
+     ((stringp (car args) )
+      (list (intern (concat (symbol-name name-symbol) "-mode")) (car args) name-symbol))
+     ((symbolp (car args))
+      (list (car args) nil name-symbol))))
    ((and (= (length args) 2)
          (symbolp (car args)))
     (list (car args) (cadr args) name-symbol))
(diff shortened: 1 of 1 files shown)


use-package.el
```diff
@@ -933,9 +933,14 @@ deferred until the prefix key sequence is pressed."
 (defun use-package-normalize/:delight (name-symbol keyword args)
   "Normalize arguments to delight."
   (cond
-   ((and (= (length args) 1)
-         (symbolp (car args)))
-    (list (car args) nil name-symbol))
+   ((= (length args) 1)
+    (cond
+     ((and (booleanp (car args)) (car args))
```
npostavs: This is just a slightly verbose way of saying `(eq (car args) t)`, right?

andersjohansson: Yeah, I guess `(eq (car args) t)` is really cleaner. I was so much into testing types when I wrote it that I just carried on...

andersjohansson:
Perhaps the error message should be changed to. And a mention of delight in the readme?

andersjohansson:
Ok. Changed the stupid check that @npostavs pointed out and changed the error message.

darkfeline:
Looks good to me, for what it's worth.

angrybakon:
Is it possible to use `delight` multiple modes in the same `use-package` clause?

I've tried:

```elisp
:delight
'((magit-status-mode "magit-status")
  (magit-commit-mode "magit-commit")
  (magit-log-mode "magit-log"))
```

But it still yields me:

```none
use-package-error: use-package: :delight expects same args as delight function
```

jwiegley:
Something like this must be done in a `:config` section.  use-package mainly tries to cover the common, repetitive use cases, but not things that are better handled with a bit of Lisp.

npostavs:
> Something like this must be done in a :config section

By "this" you mean the multi call thing that @angrybakon was trying? Or the stuff that @andersjohansson's proposed changes would allow?

Because I think the patch that @andersjohansson proposes here doesn't cover the multi call case, it just adds the same shortcuts for typical use cases to `:delight` that I added to `:diminish` (cf #20) so it makes sense to include it.

Although, looking at the `:diminish` code, it seems it might actually cover the multi call thing too...

PR #209, opened 2015-05-07
close #208

### LINKED ISSUE #208: :defer [N] example needed
I can't get the syntax of :defer [N] right, Emacs keeps complaining that :defer only takes one argument. Could you please add an example to the README?

... has use-package released a new version since `:defer [N]` was introduced? I updated my version of use-package, and now :idle is gone but :defer [N] doesn't seem to be available yet (at least in MELPA).

mcandre:
add :defer n example to the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,20 @@ this case, but that's a subtlety I'd rather avoid.
 `:defer [N]` causes the package to be loaded -- if it has not already been --
 after `N` seconds of idle time.
 
+```
+(use-package back-button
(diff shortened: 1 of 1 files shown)


jwiegley:
I recommend simplifying this example to:

``` elisp
(use-package back-button
  :commands (back-button-mode)
  :defer 2
  :init
  (setq back-button-show-toolbar-buttons nil)
  :config
  (back-button-mode 1))
```

Issue #154, opened 2015-01-24
Using 'use-package' causes error showing an appended hypen on package name


zenshinji:
Q: Why is use-package appending a hyphen on package names?

I get 'error: Package `aggressive-indent-' is unavailable' & the same thing for every package I try to set up except for 'dired-x'. 'linum-off' I was able to get working by installing it via elpa. Everything else I've tried fails in the same way as 'aggressive-indent'. I am running GNU Emacs 25.0.50.1 (x86_64-unknown-linux-gnu, GTK+ Version 3.10.8).

Thoughts?

More Detail:

When trying to use 'use-package' I run into the following error:

<code>
Warning (initialization): An error occurred while loading `/home/pjorgensen/.emacs.d/init.el':

error: Package `aggressive-indent-' is unavailable

To ensure normal operation, you should investigate and remove the
cause of the error in your initialization file.  Start Emacs with
the `--debug-init' option to view a complete error backtrace.
</code>

Here's my init.el:

<code>
;; -*- mode: emacs-lisp; coding: utf-8 -*-
;;
;; Title: init.el --- Where the magic begins
;; Author:
;; Email: 
;; Version: 1.0
;; Time-stamp: <>
;; Purpose: Set up base emacs environment


;;
;; -=[ About ]=-
;;


;; This file starts the server, loads Org-mode, loads ergoemacs, and then
;; loads the rest of our Emacs initialization from Emacs lisp embedded in
;; literate tangled Org-mode files. I try to do as little in here as
;; possible and define as much as possible in separate functional files
;; in ~/.emacs.d/ on each machine.


;;
;; -=[ Debugging Initialization ]=-
;;


;; activate debugging but disable others

(setq
 debug-on-error     t
 debug-on-signal    nil
 debug-on-quit      nil
 server-socket-dir (format "~/tmp/emacs%d" (user-uid))
 )


;;
;; -=[ Coding Initialization ]=-
;;


;; Set the coding to UTF-8 here to make sure everything after has it set.
;; Do this in as many ways as possible here to make sure it's reflected
;; properly everywhere.

(setq
 current-language-environment            "UTF-8"
 default-process-coding-system           '(utf-8 . utf-8)
 default-buffer-file-coding-system       'utf-8
 coding-system-for-read                  'utf-8
 coding-system-for-write                 'utf-8
 locale-coding-system                    'utf-8
 x-select-request-type                   '(UTF8_STRING COMPOUND_TEXT TEXT STRING)
 )
[...]

PR #210, opened 2015-05-08
Refresh package.el archives if package is missing.

thomasf:
solves #154  (which is closed).. 

I stole this snippet from Steve Purcell since I stumbled upon it when searching for something else...

https://github.com/purcell/emacs.d/blob/master/lisp/init-elpa.el#L48


thomasf:
This can probably be merged without causing problems for anyone.

thomasf:
Any specific reason for not merging this? 
The code just refreshed the package list the first time an :ensure'd  package is not found and tries again.

Issue #52, opened 2013-09-16
use-package fires twice (and :config is evaled when it shouldn't)?


Silex:
Hello,

With the following:

```
(use-package dired+
  :commands toggle-diredp-find-file-reuse-dir)

(use-package dired
  :commands (dired
             find-name-dired
             find-dired)
  :config
    (toggle-diredp-find-file-reuse-dir 1))
```

I have the following output:

```
LOADING /Users/silex/Dropbox/emacs/config/dired.el
Configuring package dired... [2 times]
Reusing Dired buffers is now ON
Configuring package dired...done (0.348s)
Reusing Dired buffers is now ON
Configuring package dired...done (0.604s)
```

Weird isn't it? This file is loaded from my .emacs... but when I `eval-buffer` it then things are only printed once to `*Message*`, yet I'm sure the file is loaded only once.

PR #217, opened 2015-06-01
Fix :config loading when package name is a file path

lionel-:
Hello,

Currently when a file is specified as package name, the `:config` block won't load. This patch fixes the issue. This is useful to load a development package without messing with `load-path`.

```lisp
(use-package ~/dir/my-package/lisp/my-package.el
  ...)
```

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -874,7 +874,7 @@ deferred until the prefix key sequence is pressed."
                (list t))))))
     (if (plist-get state :deferred)
         (unless (or (null config-body) (equal config-body '(t)))
-          `((eval-after-load ',name-symbol
+          `((eval-after-load ',(intern (file-name-base (symbol-name name-symbol)))
               ',(macroexp-progn config-body))))
       (use-package--with-elapsed-timer
           (format "Loading package %s" name-symbol)


npostavs:
This will rebreak #167.

lionel-:
hmm... it seems this is already the case without my patch. When I do the `(load-file "dired.el")` test in #52, the package gets configured multiple time (as many times as I loaded the file). This, with vanilla `use-package`

npostavs:
> the package gets configured multiple time (as many times as I loaded the file).

If you've loaded dired first, then the `:config` will run every time the `use-package` form is evaluated. This is expected (although normally you wouldn't evaluate it more than once).

The bug of #52/#167 is that the `:config` block is called when loading the config file named `dired.el` *before* loading the real dired, and then a second time *after* loading the real dired (note that the config file is only loaded once in this scenario).

lionel-:
@npostavs Thanks for the explanation. It should be ok now.

npostavs:
That looks like it should work, though "messing" with `load-path` seems simpler to me. Also, if you have a multi file package, wouldn't you *need* to mess with `load-path` anyway?

lionel-:
> "messing" with load-path seems simpler to me.

Well I had a strange situation where load-path would be altered after init. As a consequence, if the package was deferred, it would not load from the correct place. Now I finally got it right (hopefully) by calling `(package-initialize)` without setting its first argument to `t`.

> if you have a multi file package, wouldn't you need to mess with load-path anyway?

Some packages use the directory of their main file to know where to load the others. See [here](https://github.com/emacs-ess/ESS/blob/3a48dee2c35fcb66dc9fb1f40c2bde1423771fa1/lisp/ess-site.el#L86) for an example.

jwiegley:
I'm rejecting this to avoid even more custom meaning of values depending on context.

lionel-:
@jwiegley should I file a bug about `:load-path` being eagerly modified, so to speak?

`load-path` is modified at the moment of the `(use-package)` call instead of just before the package is loaded. As a result, if the package is deferred and if `load-path` is subsequently modified, the package may not get loaded from the intended path.

That was the reason I tried to load the package from an absolute path.

PR #218, opened 2015-06-01
Fix "function might not be defined at runtime" warning

YorkZ:
Hi John,

Thank you very much for this wonderful package. I just fixed a bug, let's use a simple example where foo.el is defined as:

    (require 'use-package)
    (use-package foo-lib :config (foo-lib-command))

and foo-lib.el:

    (defun foo-lib-command ())
    (provide 'foo-lib)

When byte-compiling foo.el, the compiler always produces the warning `function foo-lib-command might not be defined at runtime`.

I think the reason is because in the expanded code, the `(require foo-lib)` is nested in a `if` block preventing the compiler from evaluating it at compile time:

    (if (not (require 'foo-lib))
        ...
      )

The solution is to force the compiler evaluate the `(require 'foo-lib)` by enclosing it inside `eval-and-compile`:

    (if (not (eval-and-compile (require 'foo-lib)))
        ...
      )

York

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -882,7 +882,7 @@ deferred until the prefix key sequence is pressed."
             (use-package-concat
              (list `(require ',name-symbol))
              config-body)
-          `((if (not (require ',name-symbol nil t))
+          `((if (not (eval-and-compile (require ',name-symbol nil t)))
                 (ignore
                  (message (format "Could not load %s" ',name-symbol)))
               ,@config-body)))))))


npostavs:
I think if you don't use `(bound-and-true-p byte-compile-current-file)` then people who *don't* byte compile their init will have all their packages loaded eagerly, because compile time is the same as load time in that case.

YorkZ:
> I think if you don't use (bound-and-true-p byte-compile-current-file) then people who don't byte compile their init will have all their packages loaded eagerly, because compile time is the same as load time in that case

@npostavs: I don't understand what you meant. Can you please be more specific?

npostavs:
I looked again, your patch only applies in the non-deferred case anyway, so what you can ignore what I said.

jwiegley:
The intended fix to issues like this it to add:

``` lisp
:defines foo-lib-command
```

to your configuration block.

YorkZ:
@John:

> The intended fix to issues like this it to add: :defines foo-lib-command to your configuration block.

If you test it with my simple example you will notice that `:defines foo-lib-command` wouldn't work; `:functions foo-lib-command` wouldn't work either. Actually, the only work around that I know of is using `:commands foo-lib-command`. However, that would make the `foo-lib` autoload which may not always be what people want.


Thanks,
York

PR #225, opened 2015-06-26
Re-add support for string package names

### LINKED ISSUE #203: Problem with (use-package "foo" ...)
Hello, I suppose this issue may be counted as a feature request.

With the old use-package (pre 2.0) I could use a string as a package
name and it eventually expanded into `(eval-after-load "foo" ...)`.  But
with the current use-package, the string is interned and the use-package
clause is expanded into `(eval-after-load 'foo ...)`.

I find it inconvenient, as not every emacs file provides a feature.
For example, I used:

```elisp
(use-package "isearch"
  :diminish " 🔎"
  :config
  (setq isearch-allow-scroll t))
```

And it used to work with the old use-package, but not with the
current one, since "isearch.el" does not contain `(provide
'isearch)`.

Do you plan to add a support for string package names?

### LINKED ISSUE #244: Don't convert package names specified as strings to symbols
There exist packages which don't do `provide` and, hence, cannot be referred to with symbols, but should rather be referred to with strings.  Here is a simple example.  Consider the standard `window.el`.

```
(use-package "window"
  :defer
  :init
  (message "init")
  :config
  (message "config"))
```

The `:config` body will never be executed, and the reason is the sloppy handling of symbols vs. strings for the `name` parameter in the `use-package` macro.  The problematic code snippets are

```
(let* ((name-symbol (if (stringp name) (intern name) name)) ...
```

and

```
`((eval-after-load ',name-symbol
              ',(macroexp-progn config-body)))
```

In case of the current example (`window.el`) this will expand into

```
(eval-after-load 'window ...
```

which does nothing because there is no package loaded which would `(provide 'window)`.

To conclude, the solution is to make it expand into

```
(eval-after-load "window" ...
```

which would deliver the expected results.  Please, fix handling of the `name` parameter appropriately.

npostavs:
e.g. `(use-package "isearch")` rather than `(use-package isearch)`

fixes #203, fixes #244.

Found 2 quoting mistakes in the [initial version][1]. I think it should work now.

[1]: https://github.com/npostavs/use-package/commit/ca16fd3b1fab4eb699825b510c1b35dbde9e58f6

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -153,6 +153,23 @@ then your byte-compiled init file is as minimal as possible."
 ;; Utility functions
 ;;
 
+(defun use-package-as-symbol (string-or-symbol)
+  "If STRING-OR-SYMBOL is already a symbol, return it.  Otherwise
+convert it to a symbol and return that."
+  (if (symbolp string-or-symbol) string-or-symbol
+    (intern string-or-symbol)))
+
+(defun use-package-as-string (string-or-symbol)
+  "If STRING-OR-SYMBOL is already a string, return it.  Otherwise
+convert it to a string and return that."
+  (if (stringp string-or-symbol) string-or-symbol
+    (symbol-name string-or-symbol)))
+
+(defun use-package-load-name (name &optional noerror)
+  "Return a form which will load or require NAME depending on
+whether it's a string or symbol."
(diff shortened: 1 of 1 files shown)


jwiegley:
Is this ready to go in?  @npostavs, can you resolve the conflicts, and then I will merge it in if you think it's ready?

npostavs:
I've rebased, should be ready.

PR #234, opened 2015-07-30
Provisional hack for :bind (:map ...) support.

nickalcock:
This is almost certainly wrong: I suspect it does things in the
handler that it should do in the normalizer, and it may well not
handle various edge cases right, simply because I'm not using them
and don't know they exist.

But it seems to expand to the right code (even if this is much
less useful than I thought it was, because bind-keys doesn't create
the map if it doesn't already exist, so you have to add an :autoload
for the thing too in many cases.)

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -636,16 +636,19 @@ manually updated package."
      (concat label " wants a string, (string . symbol) or list of these")))))
 
 (defun use-package-normalize-binder (name-symbol keyword args)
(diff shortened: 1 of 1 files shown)


nickalcock:
Yep. It's wrong. Dropped. I'll test it more thoroughly next time :(

(It broke binding multiple keys at once sans a keymap.)

PR #245, opened 2015-08-17
fix quotation error

matklad:
Emacs complained about `(void-variable name)` when package failed to load, now it properly loggs missing package to \*Messages\*

The error was introduced here: https://github.com/jwiegley/use-package/commit/a863840fe666e1ae71251d5515784890bf944376#diff-7d8cebd9ba9fff18a51d6ca83a396f34L895

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -910,7 +910,7 @@ deferred until the prefix key sequence is pressed."
              config-body)
           `((if (not ,(use-package-load-name name t))
                 (ignore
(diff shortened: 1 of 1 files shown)


npostavs:
> The error was introduced here: https://github.com/jwiegley/use-package/commit/a863840fe666e1ae71251d5515784890bf944376#diff-7d8cebd9ba9fff18a51d6ca83a396f34L895

oops, I forgot to test failing scenarios.

Issue #246, opened 2015-08-30
No way to handle errors from somehow defered packages


edvorg:
Maybe provide on? For example throw en error to some handler.

jwiegley:
@edvorg What is this issue requesting?  I don't understand at all.

edvorg:
What happens if :config block generates error when package is deferred (by
keybinding, autoloader, delay, etc)? The issue is that last time I checked
there are were not way to handle this error - provide a handler to be
invoked when error happens.

jwiegley:
The error is reported to the `*Messages*` before.

edvorg:
But it's not handling, it's logging :)

PR #247, opened 2015-08-30
error handling for use-package #246

edvorg:
Hi.
I suggest a way to handle errors from packages loading process by overriding use-package handler

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -165,10 +165,10 @@ convert it to a string and return that."
   (if (stringp string-or-symbol) string-or-symbol
     (symbol-name string-or-symbol)))
 
-(defun use-package-load-name (name &optional noerror)
+(defun use-package-load-name (name)
   "Return a form which will load or require NAME depending on
 whether it's a string or symbol."
-  (if (stringp name) `(load ,name 'noerror) `(require ',name nil 'noerror)))
+  (if (stringp name) `(use-package-load ,name) `(use-package-require ',name)))
 
 (defun use-package-expand (name label form)
   "FORM is a list of forms, so `((foo))' if only `foo' is being called."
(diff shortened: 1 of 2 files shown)


use-package.el
```diff
@@ -168,7 +168,7 @@ convert it to a string and return that."
 (defun use-package-load-name (name &optional noerror)
   "Return a form which will load or require NAME depending on
 whether it's a string or symbol."
-  (if (stringp name) `(load ,name 'noerror) `(require ',name nil 'noerror)))
```
npostavs: Oh, I made a mistake in #225, those `'noerror`s should have been `,noerror`. You should probably drop the `noerror` param if you're going to ignore it anyway.

edvorg: The error goes into `use-package-require-error-handler`. So we need to omit 'noerror param here.

npostavs: Yes, I'm saying you should therefore drop the `&optional noerror` part too.

edvorg: Oh. I understood. Thank you. I will fix it.

jwiegley:
@edvorg This is not working right for me, I'm getting lots of messages like this after merging:

```
Error (use-package): helm-config :config: Wrong number of arguments: if, 1
Error (use-package): dired :config: Wrong number of arguments: if, 1
Error (use-package): bookmark :config: Wrong number of arguments: if, 1
Error (use-package): cus-edit :config: Wrong number of arguments: if, 1
Error (use-package): paredit :config: Wrong number of arguments: if, 1
Error (use-package): calfw :init: Wrong number of arguments: if, 1
```

And in `*Messages*`:

```
Eager macro-expansion failure: (wrong-number-of-arguments if 1)
File mode specification error: (wrong-number-of-arguments if 1)
```

Reverting until these can be tracked down, which I will not have time myself to do for a bit.

PR #258, opened 2015-09-23
Add support for key-chord keyword `:chords`

waymondo:


jwiegley:
@waymondo Could we implement this as an addon package that one would load after `use-package`, for example with `(require 'up-chords)`.  I'm wary of adding new core functionality that I wouldn't use (and hence couldn't maintain or determine the fitness of).  The whole promise of use-package 2.0 is that you should be able to offer an extension like this without having to add it to core use-package itself.

waymondo:
Ok that would work for me. I was pleased with how easy of an addition it was, so I tried to slip it into the main project. I will create an extension package for it then.

kaushalmodi:
@waymondo Can you post your `:chords` implementation or a link to it over here for reference? Thanks.

waymondo:
just posted it to GH (https://github.com/waymondo/use-package-chords) and submitted it to MELPA, waiting for approval.

kaushalmodi:
@waymondo :+1:

kaushalmodi:
Tagging @jordonbiondo as he might be interested in this [`use-package-chords`][1] package too.

[1]: https://github.com/waymondo/use-package-chords

jordonbiondo:
Awesome, I have been meaning to do this for myself for a long time, nice job getting it out. Now I can finally start using 2.0!

waymondo:
@jordonbiondo I was actually inspired by seeing the `:chords` keyword in your `.emacs.d`, and the 2.0 refactoring made it quite easy to port it. So props back at you!

jwiegley:
Nice and clean addon, I like it.  This will be a good example for showing others how to add more custom keywords.

PR #259, opened 2015-09-23
Pass in symbol of bind macro, for more extensible re-use of same handler

waymondo:
Since I'm going to create a small extension library that adds support for the `:chords` keyword, it would be helpful to re-use the existing `use-package-handler/:bind` method. This allows the macro to be used in the handler to be passed in as an argument to allow for this in an external package.

related to #258


waymondo:
thanks!

PR #262, opened 2015-10-02
Remove the cl dependency

Fanael:


Fanael:
Ping.

jwiegley:
Hi @Fanael.  I appreciate the offer to help, but what is the purpose of this change?  `cl` is being included at compile-time, and so can have no impact on the end user of `use-package`.

Fanael:
> what is the purpose of this change?

`cl` is deprecated in favor of `cl-lib`, so I cringe every time I see `(require 'cl)` in non-ancient code. Since cl facilities are used only in a single function, it doesn't even make sense to change to code to use `cl-lib` instead, as the function in question can be implemented using only core Emacs Lisp without any readability loss.

thomasf:
but but `cl-lib` is also a part of emacs core (whatever that means), using that is probably a good idea then.

jwiegley:
I prefer the `loop` macro, and especially `loop` to `cl-loop`. There is no problem with requiring `cl` at compile time.

PR #271, opened 2015-10-30
Miscellaneous small changes 

tarsius:
These are some suggestions. Please pick and reject as you see fit, long live Magit. Anyways...

The first commit aligns the startup messages. I find that quite helpful because it makes the initial contents of `*Messages*` much more readable.

2th removes the "Compiling package foo". I found that message to be confusing, because `use-package` is not actually compiling any packages.

3th prevents `use-package-tests.el` from being compiled and autoloaded.

### DIFF
--- a/use-package-tests.el
+++ b/use-package-tests.el
@@ -46,5 +46,7 @@
 
 ;; Local Variables:
 ;; indent-tabs-mode: nil
+;; no-byte-compile: t
+;; no-update-autoloads: t
 ;; End:
 ;;; use-package-tests.el ends here
--- a/use-package.el
+++ b/use-package.el
@@ -904,7 +904,7 @@ deferred until the prefix key sequence is pressed."
           `((eval-after-load ,(if (symbolp name) `',name name)
               ',(macroexp-progn config-body))))
       (use-package--with-elapsed-timer
-          (format "Loading package %s" name)
+          (format "    Loading package %s" name)
         (if use-package-expand-minimally
             (use-package-concat
              (list (use-package-load-name name))
--- a/use-package.el
+++ b/use-package.el
@@ -1052,8 +1052,6 @@ this file.  Usage:
                               (plist-get args* :defines))
                     (with-demoted-errors
                         ,(format "Cannot load %s: %%S" name)
-                      ,(if use-package-verbose
-                           `(message "Compiling package %s" ',name-symbol))
                       ,(unless (plist-get args* :no-require)
                          (use-package-load-name name)))))))
 


jwiegley:
@tarsius The reason I put the "Compiling package" text in is so that I can locate the package responsible for certain compilation errors.  I found it was impossible to track them down without it.

tarsius:
I haven't experienced any errors in my use of `use-package`, but I guess it might be a different story when you are *working on* `use-package`. It's perfectly fine with me if you don't pick this (as it is for the other commits), I'll just patch this locally (I use submodules, so that is trivial).

vyp:
Sorry for being completely off topic, @tarsius how do you patch submodules? (if you have a link or something explaining it)

tarsius:
@vyp If been doing it manually for a few years and have recently started (again) to automate the process a bit. I intend to release a package at some point, but for now, let me paraphrase what I do:

* Installing a package: `git submodule add -name foo git://github.com/bar/foo.git lib/foo`
* Installing dependencies: look at the code, then install what is needed (I look at the code anyway)
* Compile, create autoload files, generate info files: `cd ~/.emacs.d` then `make`, `make quick`, or `make lib/foo`. Sorry the code ain't ready to be released.
* Updating packages:

   To be able to easily update, this setup is required:

```lisp
(magit-add-section-hook 'magit-status-sections-hook
                        'magit-insert-unpulled-module-commits
                        'magit-insert-unpulled-commits)
 ```

* ... then ...
  * Show status for `~/.emacs.d`.
  * Fetch submodules: `f m`.
  * Visit submodule status: `RET`.
  * Pull changes: `P P`, or not.
  * Recompile: `make`, `make quick`, or `make lib/foo lib/bar`.

* Maintain local patches.
  * Create commits on `master`, prefix commit messages with e.g. "[local] "
  * Configure to rebase instead of merge: `M-x magit-branch-toggle-rebase` (from `magit-rockstar.el`).
  * Pull. If necessary resolve conflicts.

vyp:
That's great, thanks for taking the time to elaborate, really appreciated! I don't want to spam here with off topic messages, but just want to say a few things.

> Sorry the code ain't ready to be released.

That's okay, more curious than anything. And I use quelpa to install 'local' packages anyway, and that does seem to generate autoloads (probably via package-build.el or package.el), although I'm not sure if it generates info files.

> Configure to rebase instead of merge

Nice, this is clever. But say we want to put our emacs config on another machine, how will these `[local]` commits be transferred? (unless you don't really care about that) My best guess would be to create a new repository for the package that we can push to (i.e. 'fork'), and just use that (i.e. pull from an 'upstream' remote). Or maybe somehow maintain a set of .patch files or something..? (Or perhaps instead of using git, you just transfer your config via rsync or other normal means?) Anyway, just wondering if you have already thought of this and have some clever solution.

jwiegley:
Included some of the changes manually.

tarsius:
> The reason I put the "Compiling package" text in is so that I can locate the package responsible for certain compilation errors. I found it was impossible to track them down without it.

Maybe that message could be printed only when `(eq use-package-verbose 'debug)`?

jwiegley:
That's totally fair.

jwiegley:
Done.

tarsius:
Great, thanks!

Issue #260, opened 2015-09-29
Entering "(use-package kill" causes Emacs frozen


goofansu:
Emacs version: `GNU Emacs 24.5.1 (x86_64-apple-darwin13.4.0, NS apple-appkit-1265.21) of 2015-04-11 on builder10-9.porkrind.org`

Problem:

I entered `(use-package kill` by accident, but I found every time I enter this sentence, Emacs calls melpa first and then frozen.

My config to reproduce this problem:

```
(require 'package)
(setq package-enable-at-startup nil)
(add-to-list 'package-archives
             '("melpa" . "http://melpa.org/packages/"))

(package-initialize)

;; Bootstrap `use-package'
(unless (package-installed-p 'use-package)
  (package-refresh-contents)
  (package-install 'use-package))

(eval-when-compile
  (require 'use-package))
(require 'diminish)                ;; if you use :diminish
(require 'bind-key)                ;; if you use any :bind variant

(setq use-package-always-ensure t)

(use-package swiper
  :bind (("C-s" . swiper)
	 ("C-r" . swiper)
	 ("s-z" . ivy-resume))
  :init
  (setq ivy-use-virtual-buffers t)
  (setq ivy-re-builders-alist
	'((t . ivy--regex-fuzzy)))
  :config (ivy-mode 1))

(use-package counsel
  :bind (("s-g" . counsel-git-grep)
	 ("s-x" . counsel-M-x)
	 ("s-y" . counsel-yank-pop)))

(use-package projectile
  :diminish projectile-mode
  :bind (("s-p" . projectile-switch-project)
	 ("s-f" . projectile-find-file)
	 ("s-d" . projectile-find-dir)
	 ("s-c" . projectile-compile-project)
	 ("s-t" . projectile-test-project))
  :init (setq projectile-completion-system 'ivy)
  :config (projectile-global-mode))

(use-package company
  :diminish company-mode
  :init
  (setq company-idle-delay 0.5)
  (setq company-tooltip-limit 10)
  (setq company-minimum-prefix-length 2)
  (setq company-tooltip-flip-when-above t)
  :config (global-company-mode))

''' just enter "(use-package kill" here
```

thomasf:
I don't think this is an issue with use-package.

goofansu:
@thomasf Only when entering `use-package` with words like `kill` or `recentf` will cause this problem.

npostavs:
[...]

Issue #190, opened 2015-03-23
Issue with using the `use-package-always-ensure` option


kaushalmodi:
Hi , 

I tried out the new option and set it to `t`.

But it gives me this error:

> Eager macro-expansion failure: (error "Package `defuns-' is unavailable")
> package-compute-transaction: Package`defuns-' is unavailable

I have the below in my emacs init:

``` el
(defun prepend-path ( my-path )
  (setq load-path (cons (expand-file-name my-path) load-path)))

(prepend-path (concat user-emacs-directory "/elisp"))
;; ...
(package-initialize)
;; ...
(eval-when-compile
  (require 'use-package)
  (setq use-package-always-ensure t))
(use-package bind-key)
(use-package defuns)
```

The [`defuns.el` is present in that `elisp/` path](https://github.com/kaushalmodi/.emacs.d/blob/master/elisp/defuns.el).

But is looks like `defuns` is being tried to be installed by the package manager (even though it is available in the `load-path`) when this new option is `t` (or probably because `:ensure` is `t`).

jwiegley:
Can you run `M-x toggle-debug-on-error` and paste the full backtrace of the error?  Also can you show me the evaluation of:

``` elisp
(pp-to-string (macroexpand '(use-package defuns)))
```

kaushalmodi:
Sure.

Here is the output of `emacs --debug-init`

```el
Debugger entered--Lisp error: (error "Package `defuns-' is unavailable")
  signal(error ("Package `defuns-' is unavailable"))
  error("Package `%s-%s' is unavailable" defuns "")
  package-compute-transaction(nil ((defuns)))
  package-install(defuns)
  (progn (package-install package))
  (if (not (package-installed-p package)) (progn (package-install package)))
  use-package-ensure-elpa(defuns)
  (progn (require (quote package)) (use-package-ensure-elpa package-name))
  (if package-name (progn (require (quote package)) (use-package-ensure-elpa package-name)))
  (let ((package-name (or (and (eq ensure t) name-symbol) ensure))) (if package-name (progn (require (quote package)) (use-package-ensure-elpa package-name))))
  (let ((body (use-package-process-keywords name-symbol rest state))) (let ((package-name (or (and (eq ensure t) name-symbol) ensure))) (if package-name (progn (require (quote package)) (use-package-ensure-elpa package-name)))) body)
  use-package-handler/:ensure(defuns :ensure t (:config (t)) nil)
[...]

PR #273, opened 2015-11-03
Move :ensure option execution from macro expansion phase to runtime phase

kovrik:
**Description:**
Currently `:ensure` option is executed during macro expansion time.

This results in issues like:
https://github.com/jwiegley/use-package/issues/260 and https://github.com/company-mode/company-mode/issues/387

**Issue:**
If you have `company-mode` installed, `use-package` installed and `use-package-always-ensure` set to `t`, then typing `(use-package 123|` will make `company-mode` macro-expand `use-package` macro which will trigger contacting package repositories and fetching `123` package.

**Solution:**
This pull request moves execution of `:ensure` option from macro expansion time to runtime.

**WARNING**
**Please, double-check and test everything!**
Macros are tricky things, I might forgot to add a backquote or something.

I have tested several cases locally and for me it worked as expected, solved the issue.
But I'm not sure if this works in all cases.

**Also**, note that using `company-mode` + `use-package` is still pretty slow because `use-package` is a huge macro and `company-mode` expands it on each auto-completion attempt.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -463,13 +463,11 @@ manually updated package."
 
 (defun use-package-handler/:ensure (name keyword ensure rest state)
   (let ((body (use-package-process-keywords name rest state)))
-    ;; This happens at macro expansion time, not when the expanded code is
-    ;; compiled or evaluated.
-    (let ((package-name (or (and (eq ensure t) (use-package-as-symbol name)) ensure)))
-      (when package-name
-        (require 'package)
-        (use-package-ensure-elpa package-name)))
-    body))
+    `((let ((package-name (or (and (eq ,ensure t) (use-package-as-symbol ',name)) ,ensure)))
+          (when package-name
+            (require 'package)
+            (use-package-ensure-elpa package-name)))
+        ,@body))) 
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;


jwiegley:
Since I don't use `:ensure` at all, can a few people who do please test this and verify that it does not break that support for them?

use-package.el
```diff
@@ -463,13 +463,11 @@ manually updated package."
...
 (defun use-package-handler/:ensure (name keyword ensure rest state)
   (let ((body (use-package-process-keywords name rest state)))
-    ;; This happens at macro expansion time, not when the expanded code is
-    ;; compiled or evaluated.
-    (let ((package-name (or (and (eq ensure t) (use-package-as-symbol name)) ensure)))
-      (when package-name
-        (require 'package)
-        (use-package-ensure-elpa package-name)))
-    body))
+    `((let ((package-name (or (and (eq ,ensure t) (use-package-as-symbol ',name)) ,ensure)))
```
npostavs: You should probably use `make-symbol` for the `package-name` binding.

Also, I *think* it should be `',(use-package-as-symbol name)`.

kovrik: @npostavs 

>Also, I think it should be ',(use-package-as-symbol name).

Hmm, I think it shouldn't.
If you do ` ',(use-package-as-symbol name)`, then it will be executed on each macro expansion. `use-package-as-symbol` interns symbols, which means that if you type `(use-package test123|` and do little pause between each letter, then it will call `use-package-as-symbol` many times and will intern symbols `t`, `te`, `tes`, `test` and so on.

Correct me if I'm wrong.

>You should probably use make-symbol for the package-name binding.

Not sure about this one as well.
`name` is a symbol, not a string. 
`make-symbol` requires a string.

How do you suggest to use `make-symbol` here?
Convert `name` (already interned by `use-package-as-symbol`) to a string and then `make-symbol` out of this string?

npostavs: > which means that if you type (use-package test123| and do little pause between each letter, then it will call use-package-as-symbol many times and will intern symbols t, te, tes, test and so on.

Well, only in the case where you type `(use-package "test123`, but still a good point.

> `name` is a symbol, not a string.

No, I was talking about `package-name` (but FYI, `name` is *usually* symbol, sometimes a string). As in
```elisp
(let ((package-name-var (make-symbol "package-name")))
  `(let ((,package-name-var (or (and (eq ,ensure t) ...))))))
```
But actually, looking again, the `,@body` is ouside of `package-name`'s scope, so maybe it's fine to leave it as a normal symbol.

wasamasa: @kovrik Sorry, I don't use use-package.

kaushalmodi:
I believe this solves https://github.com/jwiegley/use-package/issues/190?

kovrik:
@kaushalmodi 
Not sure.
Probably it will solve eager macro expansion issue, but I think it won't fix things you mentioned:

>:ensure t must not try to install a package using package manager if one of the below is true:
> - User specified :load-path for that package
> - That package already exists on load-path

My fix just moves evaluation of `:ensure` to runtime.
But in runtime it still will try to execute the following function:

`(use-package-ensure-elpa package-name)`

Which I believe will do the same thing it does now.
I'd suggest taking a look at this function.
I'll do that a little bit later.

jwiegley:
@kovrik Hi, I saw your comment on Reddit, my apologies for the frustration.

My only worry is that because makes a significant change to `:ensure`, I'd like at least one other person to confirm that it doesn't break their configuration. Then we can merge it.

@thomasf Could you test this?

kovrik:
@jwiegley 
Hi!
Sorry if I sounded harsh, didn't mean that.

I understand your worries, moving the whole `:ensure` option from macro-expansion to runtime is a big change, that's why I asked other people to test all possible cases.

I'll try to ask other guys to take a look and review changes.

Thank you!

zakkak:
It doesn't seem to break the functionality of `use-package` but it breaks `company-mode` for `use-package`. That is, it does not provide packages as completions but variables.

kovrik:
@zakkak 
>it does not provide packages as completions but variables.

Can you explain please?

Just tried old version and new version - can't see any package completions.
The only difference I see is that new version is usable with `use-package-always-ensure t`

zakkak:
With the original version when I type `(use-package mon` company-mode starts auto-completion and provides a list of packages starting with mon as candidates.  Then when I type another letter, e.g. o, it narrows the list to packages starting with mono.  Every time I type a letter the package-list gets refreshed (resulting in huge delay).  After applying your patch the auto-completion is instant but the completion candidates are variables and not packages from the package list.

kovrik:
@zakkak 
Can you please provide your `company-mode` settings?

It seems I don't use that back-end to complete package names.

zakkak:
The settings:

```
(use-package company
  :config (setq company-idle-delay 0))
(add-hook 'after-init-hook 'global-company-mode)
```

The `company-backends` variable:

```
company-backends is a variable defined in `company.el'.
Its value is (company-bbdb company-nxml company-css company-eclim company-semantic company-clang company-xcode company-cmake company-capf
	      (company-dabbrev-code company-gtags company-etags company-keywords)
	      company-oddmuse company-files company-dabbrev)
```

npostavs:
> With the original version when I type (use-package mon company-mode starts auto-completion and provides a list of packages starting with mon as candidates.

I can't reproduce this, with latest `master` of use-package (0f092288c6f2ee15ea746409c121e2863cc9e457) and company-mode (company-mode/company-mode@c656b25f9e987799bf6b07df761561fb96dda6c2). I get variables and functions as completion candidates.

kovrik:
@npostavs 
Same for me.
Can't get list of packages in company-mode.

@zakkak 
Could you please show us a screenshot?

zakkak:
You are both right!  I don't know why I got that impression, probably due to some package providing a variable with the same name.

So I guess the patch seems OK to me ;)

kovrik:
@zakkak 
Yay! Glad to hear :)
Thank you!

kovrik:
Soo... :)
Any updates? Anybody else wants to test it?

kovrik:
@jwiegley 
Thank you, John!

thomasf:
This change seems to have brought bugs.. see my comment on https://github.com/jwiegley/use-package/commit/a591cd244495b9e8dea8b1ebd5eb52a096912b5a

thomasf:
sorry for not paying attention to this issue, I see that I was asked to verify this a week ago or so.. I've been very busy lately, sorry.

PR #276, opened 2015-11-05
Support for after keyword.

### LINKED ISSUE #274: eval-after-load
I can't see any good way of doing the equivalent of "eval-after-load". I am thinking, for instance, of adding a org-mode exporter, which can't be autoload but needs to be installed as soon as org-mode has loaded.

phillord:
Turned out to be relatively simple. I like the reworked code base!

After provides `eval-after-load' functionality for any number of
features. This helps to avoid deeply nested `use-package' statements.

Closes #274

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -126,6 +126,7 @@ the user specified."
     :defines
     :functions
     :defer
+    :after
     :demand
     :init
     :config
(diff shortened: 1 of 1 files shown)


vyp:
Great! Imo you should also add a note somewhere in the README about it.

zakkak:
Nice! 

I find "feature" misleading, though.  Is there any specific reason you prefer it over "package"?

manuel-uberti:
One newbie question: does `:after` work as a replacement for `with-eval-after-load` as well? I mean, I got this in my init.el:
```
:init (with-eval-after-load 'yasnippet
          (bind-key "C-c h y" #'helm-yas-complete))
```
Can I use `:after` for such cases?

manuel-uberti:
Ok, I had a look at the code. I guess the answer is no, so nevermind. :-)

manuel-uberti:
Ok, the answer is yes:
```
  :after yasnippet
  :init (bind-key "C-c h y" #'helm-yas-complete)
```

phillord:
@zakkak "Feature" is strictly correct, rather than "package". A package can contain several files. Each file with `(provide 'feature)` form at the end is a feature.

I've just copied it from the `eval-after-load` documentation. Probably it should support files (as eval-after-load does), but the current implementation doesn't.

zakkak:
@phillord thanks for the clarification.  So in this case, `use-package` seems misleading and should become `use-feature` and then maybe have a `:provided` field to indicate the package that provides the corresponding feature :)

phillord:
@zakkak Sounds like a good plan to me!

justmytwospence:
> Defer loading of a package until after any of the named features are loaded.

Does this really mean "any"? Or should it be "all"?

phillord:
As implemented, it is any.

PR #278, opened 2015-11-09
Handle case where keymap is a variable

nandryshak:
As seen in this issue: https://github.com/waymondo/use-package-chords/issues/3

```
(defun test-bind (key command &optional keymap)
  (bind-key key command keymap))

(test-bind "M-B" 'eww)
```

This makes `personal-keybindings` look like this:

```
((("M-B" . keymap) eww eww))
```

which throws an error when you do `M-x describe-personal-keybinds`. That `keymap` symbol is not what we want, we want this:

```
((("M-B") eww eww))
```

Which is what this pull request will fix. 

If there is a more idiomatic solution, feel free to use it.

### DIFF
--- a/bind-key.el
+++ b/bind-key.el
@@ -151,7 +151,8 @@ spelled-out keystrokes, e.g., \"C-c C-z\". See documentation of
                        (read-kbd-macro ,namevar)))
             (,kdescvar (cons (if (stringp ,namevar) ,namevar
                                (key-description ,namevar))
-                             (quote ,keymap)))
+                             (when ,keymap
+                               (quote ,keymap))))
             (,bindingvar (lookup-key (or ,keymap global-map)
                                      ,keyvar)))
        (add-to-list 'personal-keybindings


nandryshak:
Actually this doesn't fix it completely yet. I will update this if I find a solution.

npostavs:
> If there is a more idiomatic solution, feel free to use it.

I think the easiest way to fix it will be to convert `bind-key` to a function. It's something I considered doing a while back, but avoided since I wasn't sure why it's a macro in the first place, and I didn't want to break things. But now it appears it's broken anyway, so...

jwiegley:
Looking at it now, I'm not entirely sure why it was a macro either...

nandryshak:
Well, if you're passing in a keymap, you want to be able to add the name of the map to `personal-keybindings`, which you can't do with a function, right? 

Regardless, I think this problem might be better solved by the author of `use-package-chords`, by rewriting one of his functions as a macro.

npostavs:
> Well, if you're passing in a keymap, you want to be able to add the name of the map to personal-keybindings, which you can't do with a function, right?

Hmm, you'd have to quote it, which would be a non-backwards compatible change.

nandryshak:
Right. And then the quoted symbol would cause problems for the `define-key` portion. I guess you could `eval` it. But if you're using `eval`, you probably should be using a macro instead, right? And now we've come full circle.

npostavs:
> And then the quoted symbol would cause problems for the define-key portion.

Nope. A symbol that is fbound to a keymap *is* a keymap.
```
keymapp is a built-in function in `C source code'.

(keymapp OBJECT)

Return t if OBJECT is a keymap.

A keymap is a list (keymap . ALIST),
or a symbol whose function definition is itself a keymap.
```

nandryshak:
(defun test-bind-key (key command &optional keymap)
      (define-key keymap key command))
    (test-bind-key (kbd "M-B") 'eww 'global-map)

Throws an error for me: `test-bind-key: Wrong type argument: keymapp, global-map`

jwiegley:
Don't quote `global-map'.

npostavs:
> Throws an error for me: test-bind-key: Wrong type argument: keymapp, global-map

Oh, it's `bound`, not `fbound`.

> I guess you could eval it. But if you're using eval, you probably should be using a macro instead, right?

We could use `symbol-value` if given symbols that don't satisfy `keymapp`.

jwiegley:
@nandryshak What is the status on this PR?

PR #280, opened 2015-11-10
use-package-ensure-elpa: quote ensure

### LINKED ISSUE #279: Symbol's value as variable is void: auctex
I'm a newbie in emacs and auctex.
I'm using Emacs 23.4 (9.0) in Mac OS X 10.7.5

I was trying to follow the answer in 
http://tex.stackexchange.com/questions/239544/adding-keybord-shortcut-for-auctex-in-emacs

I don't know how to do things in a right way or in an efficient way, so in my .emacs file I added

(load-file "~/.emacs.d/site-lisp/macroexp.el")
(load-file "~/.emacs.d/site-lisp/cl-lib.el")
(load-file "~/.emacs.d/site-lisp/bind-key.el")
(load-file "~/.emacs.d/site-lisp/use-package.el")

after downloading those files and copying to my directory site-lisp
then as a test, I added

(use-package latex
  :ensure auctex
  :config
  (add-to-list 'LaTeX-font-list
               '(?\C-p "\\pname{" "}")))

Then when I run emacs, it says

Symbol's value as variable is void: auctex

Note that the above lines that I added to my .emacs are added after

(load "auctex.el" nil t t)

Can anyone tell me what the problem is?

npostavs:
fixes #279.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -464,7 +464,7 @@ manually updated package."
 
 (defun use-package-handler/:ensure (name keyword ensure rest state)
   (let ((body (use-package-process-keywords name rest state)))
-    `((let ((package-name (or (and (eq ,ensure t) (use-package-as-symbol ',name)) ,ensure)))
+    `((let ((package-name (or (and (eq ',ensure t) (use-package-as-symbol ',name)) ',ensure)))
           (when package-name
             (require 'package)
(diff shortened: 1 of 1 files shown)


npostavs:
hah, too slow

PR #281, opened 2015-11-11
{un,}bind-key: implement by expanding to function

npostavs:
This allows fixing #278 by doing

```elisp
(defun test-bind (key command &optional keymap)
  (bind-key-fun key command keymap))

(test-bind "M-B" 'eww)
```

Although possibly it's better to rewrite `test-bind` as a macro so it can autoquote its `keymap` argument too (cf waymondo/use-package-chords#4), I think it's worth rewriting `bind-key` this way since it's clearer what it does and what it uses the `defmacro` for.

### DIFF
--- a/bind-key.el
+++ b/bind-key.el
@@ -142,36 +142,37 @@ KEY-NAME may be a vector, in which case it is passed straight to
 `define-key'. Or it may be a string to be interpreted as
 spelled-out keystrokes, e.g., \"C-c C-z\". See documentation of
 `edmacro-mode' for details."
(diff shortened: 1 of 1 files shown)


npostavs:
Hmm, I think this may have the same problem what was raised here https://github.com/jwiegley/use-package/commit/a591cd244495b9e8dea8b1ebd5eb52a096912b5a#commitcomment-14306912

jwiegley:
I think I'll pass on this change until it becomes a more serious issue.

Issue #299, opened 2015-12-26
flycheck reported fail for `:pin` also the archive was in package-list


Gnouc:
Here is part of my configuration:
```
(require 'package)
(setq package-archives '(("gnu" . "https://elpa.gnu.org/packages/")
                         ("marmalade" . "https://marmalade-repo.org/packages/")
                         ("melpa" . "https://melpa.org/packages/")
                         ("melpa-stable" . "http://melpa.milkbox.net/packages/")
                         ("elpy" . "https://jorgenschaefer.github.io/packages/")))

;; markdown-mode
(use-package markdown-mode
  :ensure t
  :pin marmalade
  :mode "\\.md\\'"
  :config
  (add-hook 'markdown-mode-hook
            (lambda ()
              (when buffer-file-name
                (add-hook 'after-save-hook
                          'check-parens
                          nil t)))))
```

Although I could installed `markdown-mode` successfully, every time I save the buffer, `flycheck` with `emacs-lisp` as default checker always reported:
```
Archive 'marmalade` requested for package 'markdown-mode' is not available
```

I also replaced `marmalade` with `gnu` or `melpa-stable` with no luck.

What cause this issue and how to fix?

Gnouc:
@jwiegley Could you please take a look at this issue?

npostavs:
Does it work if you wrap the `setq` in `eval-when-compile`? I think it might be checking the the archives at macro-expansion time, before your `setq` has taken effect.

Gnouc:
@npostavs Yes, wrapping `setq` in `eval-when-compile` made the error from `:pin` disappear, but made a bunch of other warning messages, example my [sh hook](https://github.com/Gnouc/emacs.d/blob/master/init.el#L443) was warned with message
```
assignment to free variable `sh-basic-offset'
assignment to free variable `sh-indentation'
```

Gnouc:
@npostavs Yes, wrapping `setq` in `eval-when-compile` made the error message disappear.

npostavs:
So currently we have
```elisp
(defun use-package-handler/:pin (name keyword archive-name rest state)
...
    ;; This happens at macro expansion time, not when the expanded code is
    ;; compiled or evaluated.
...
```

But actually, I think the `:pin` should take effect the same time as `:ensure`: runtime and during byte-compilation. @kovrik, @thomasf  does that seem right to you?

jwiegley:
I don't use `:pin` at all, so I have nothing to add to this discussion.

[...]

PR #302, opened 2016-01-05
Move :pin out of macro expansion phase

kovrik:
Fix (but I'm not sure) for #299

Changes:

1) move `:pin` option out of macro-expansion phase 
(the same thing we did for the `:ensure` option)

2) fix order of `:pin` and `:ensure` options.
`use-package-sort-keywords` function works correctly and puts `:pin` before `:ensure` after sorting.
But, after macro-expansion `:pin`'s body appears after the `:ensure`s body.
So, it pins the package, but after the installation (which is obviously wrong).

3) replace 
`(push (cons package archive-name) package-pinned-packages)` 
with 
`(add-to-list 'package-pinned-packages (cons package archive-name) t)`
Because `push` doesn't check if entry is already a member of a list. 
Hence, if you have something like `(use-package test :pin melpa)` and eval that form N times, you will get N '(test . "melpa")` entries in your `package-pinned-packages` list.
`add-to-list` does the same thing as `push`, but also checks if entry is already there.

After this changes I have:
1) it inserts new entry into `package-pinned-packages` only once (no more bloating!)
2) if I `pin` to the wrong archive (which is not in `package-archives`), I get the error (`"Archive '%s' requested for package '%s' is not available."`)
3) if I `pin` to the correct archive, then it installs required package using that archive
4) if I `byte-compile`, then it correctly `pin`s and installs package

Example:
``` lisp
(pp-macroexpand-expression '(use-package markdown-mode :ensure t :pin melpa))
```
Gives:
``` lisp
(progn
  (use-package-pin-package 'markdown-mode "melpa")
  (progn
    (require 'package)
    (use-package-ensure-elpa 'markdown-mode))
  (let
      ((now ...))
    (message "%s..." "Loading package markdown-mode")
    (prog1
        (if ... ...)
      (let ... ...))))
```

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -419,23 +419,23 @@ manually updated package."
   (let ((archive-symbol (if (symbolp archive) archive (intern archive)))
         (archive-name   (if (stringp archive) archive (symbol-name archive))))
     (if (use-package--archive-exists-p archive-symbol)
-        (push (cons package archive-name) package-pinned-packages)
+        (add-to-list 'package-pinned-packages (cons package archive-name) t)
       (error "Archive '%s' requested for package '%s' is not available."
              archive-name package))
     (package-initialize t)))
 
 (defun use-package-handler/:pin (name keyword archive-name rest state)
-  (let ((body (use-package-process-keywords name rest state)))
-    ;; This happens at macro expansion time, not when the expanded code is
-    ;; compiled or evaluated.
-    (if (null archive-name)
-        body
-      (use-package-pin-package name archive-name)
-      (use-package-concat
-       body
-       `((push '(,(use-package-as-symbol name) . ,archive-name)
-               package-pinned-packages)
-         t)))))
+  (let ((body (use-package-process-keywords name rest state))
+        (pin-form (if archive-name
+                      `(use-package-pin-package ',name ,archive-name))))
+    ;; We want to avoid pinning packages when the `use-package'
+    ;; macro is being macro-expanded by elisp completion (see
+    ;; `lisp--local-variables'), but still do pin packages when
+    ;; byte-compiling to avoid requiring `package' at runtime.
+    (if (bound-and-true-p byte-compile-current-file)
+        (eval pin-form)              ; Eval when byte-compiling,
+      (push pin-form body))          ; or else wait until runtime.
+    body))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;
--- a/use-package.el
+++ b/use-package.el
@@ -419,7 +419,7 @@ manually updated package."
   (let ((archive-symbol (if (symbolp archive) archive (intern archive)))
         (archive-name   (if (stringp archive) archive (symbol-name archive))))
     (if (use-package--archive-exists-p archive-symbol)
-        (add-to-list 'package-pinned-packages (cons package archive-name) t)
+        (add-to-list 'package-pinned-packages (cons package archive-name))
       (error "Archive '%s' requested for package '%s' is not available."
              archive-name package))
     (package-initialize t)))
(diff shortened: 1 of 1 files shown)


kovrik:
@npostavs @jwiegley @thomasf 
Could you please review this pull-request?

I did the same thing we did for the `:ensure` option.
Did some tests locally - worked good for me (but I'm stupid, so, please, test it yourself).

I hope I didn't break other stuff :)

I can see the following scenarios to be tested:
- normal usage of `use-package` with `:pin` and `:ensure`. Just eval the form.
- install during byte-compilation

I have tested my local config + @Gnouc's from #299.
For me it worked well.

use-package.el
```diff
@@ -419,23 +419,23 @@ manually updated package."
   (let ((archive-symbol (if (symbolp archive) archive (intern archive)))
         (archive-name   (if (stringp archive) archive (symbol-name archive))))
     (if (use-package--archive-exists-p archive-symbol)
-        (push (cons package archive-name) package-pinned-packages)
+        (add-to-list 'package-pinned-packages (cons package archive-name) t)
```
npostavs: Why pass `t` for `APPEND`?

use-package.el
```diff
@@ -419,23 +419,23 @@ manually updated package."
...
-        body
-      (use-package-pin-package name archive-name)
-      (use-package-concat
-       body
-       `((push '(,(use-package-as-symbol name) . ,archive-name)
-               package-pinned-packages)
-         t)))))
+  (let ((body (use-package-process-keywords name rest state))
+        (pin-form (if archive-name
+                      `(use-package-pin-package ',name ,archive-name))))
```
npostavs: I think you should have `',(use-package-as-symbol name)` to be safe.

use-package.el
kovrik: Well, actually, I'm not sure if this change is safe.
`push` is a macro, where `add-to-list` is a function.
`add-to-list` docs say:

>This is handy to add some elements to configuration variables,
but please do not abuse it in Elisp code, where you are usually
better off using `push' or `cl-pushnew'.

What do you think?

Maybe replace it with `member` check + `push`?

use-package.el
kovrik: Good point.
Will do that.

use-package.el
```diff
@@ -419,23 +419,23 @@ manually updated package."
...
+    ;; We want to avoid pinning packages when the `use-package'
+    ;; macro is being macro-expanded by elisp completion (see
+    ;; `lisp--local-variables'), but still do pin packages when
+    ;; byte-compiling to avoid requiring `package' at runtime.
```
npostavs: I think it would be better to reference the `:ensure` behaviour instead of repeating its comment, something like:
``;; Pinning should occur just before ensuring, see `use-package-handler/:ensure'.``

kovrik: Will do that. Thanks!

use-package.el
npostavs: It should be safe. It has problems with local lexically bound variables, but there is no problem with globals. And `t` for `APPEND` wouldn't cause a problem, it just seems unneeded.

`unless member push` would be fine too, though a bit more verbose. `cl-pushnew` would be perfect, except it has the annoyance of requiring `cl-lib` at runtime (its expansion uses `cl-adjoin`).

kovrik:
@npostavs 
Pushed fixes

npostavs:
Looks good to me.

kovrik:
Thank you very much for the review!
Would be awesome if you could do some tests :)

kovrik:
@Gnouc confirms that this pull-request fixes #299

kovrik:
@jwiegley @thomasf @npostavs 
Terribly sorry, but any update on this?

thomasf:
I have not had the time to check this out yet, sorry.

thomasf:
this seems to work fine for. Obviously pinning did not work before I tried using this patch and now it does.

kovrik:
@thomasf
Thank you! :+1: 

@jwiegley
Seems people are happy, can we merge it?

adouzzy:
Quick question:
Earlier if I use too many pin:, startup time would become very long.
Would this pull solve this problem ?

thomasf:
@adouzzy that's strange since the only thing :pin does is to set a variable. It should not affect speed anywhere.. Unless package.el does some computations everytime that variable changes or something.

kovrik:
@adouzzy 
I agree with Thomas: can't imagine how `:pin` can affect startup time.
But I will take a look into the code and do some tests.

adouzzy:
Thanks @thomasf @kovrik. 
In the readme, it writes
"NOTE: if you pin a lot of packages, it will be slightly slower to start Emacs compared to manually adding all packages to the package-pinned-packages variable. However, should you do it this way, you need to keep track of when (package-initialize) is called, so letting use-package handle it for you is arguably worth the cost."
And I found it to be true in my case. I have more than 40 :pin and moves them into the package-pinned-packages variable. This solves my problem. I am just curious here.

kovrik:
@adouzzy 
Hm, I will take a look, maybe we can improve `:pin` somehow.
Can you please share your `init.el`?

PR #303, opened 2016-01-08
Allow string values in cons for :bind keywords

waymondo:
It is possible with `bind-key` and `define-key` (and also `bind-chord`
and `key-chord-define`) to define a binding to a string's value, i.e:

``` elisp
(bind-key "C-;" "the ")
(bind-chord "^^" "λ")
```

This adds an option for `(use-package-normalize-pairs)` that allows
string values to be given with the `:bind` (and also `:chord`) keywords
to expand into these definitions.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -644,8 +644,14 @@ manually updated package."
            (and allow-vector (vectorp (car x))))
        (symbolp (cdr x))))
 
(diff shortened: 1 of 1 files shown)


waymondo:
@jwiegley does this seem like a valid inclusion? do you want to run it by anyone else?

PR #314, opened 2016-02-04
When :ensure is used install package as a selected package.

thierryvolpiatto:
Also shutup bytecompiler about package-archive-contents.

* use-package.el (use-package-ensure-elpa): Add package to selected package
by using second arg of package install.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -462,7 +462,9 @@ manually updated package."
   (if (package-installed-p package)
       t
     (if (or (assoc package package-archive-contents) no-refresh)
-        (package-install package t)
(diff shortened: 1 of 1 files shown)


syohex:
Second argument of `package-install` was introduced at Emacs 25. It causes error on older Emacs(<= 24.5).

thierryvolpiatto:
Syohei YOSHIDA <notifications@github.com> writes:

> Second argument of package-install was introduced at Emacs 25. It
> causes error on older Emacs(<= 24.5).

Indeed thanks (I am using an enhanced version of package even on 24.5).

-- 
Thierry

PR #316, opened 2016-02-06
Upper casing Cs corresponding to Ctrl

dudebout:
A number of Cs corresponding to Ctrl have been lower cased in comments in cd65aebaa.


jwiegley:
@dudebout Thank you!

PR #342, opened 2016-04-24
Some minor fixes

### LINKED ISSUE #341: use-package-always-defer duplicates some binding autoloads
Consider two use-package forms:

```elisp
(use-package test
  :bind (("C-c" . flargle)))

(use-package example
  :bind (("Moo" . bagel)))
```

After evaluating the first form, with `use-package-always-defer` set to `nil`, the second form expands in part to:

```elisp
(progn (unless (fboundp (quote bagel)) (autoload (function bagel) "example" nil t)) (ignore (let* (... ... ... ...) (add-to-list ... ...) (define-key ... key ...))))
```

while with `use-package-always-defer` set to `t`, it expands partially to:

```elisp
(progn (unless (fboundp (quote bagel)) (autoload (function bagel) "example" nil t)) (unless (fboundp (quote flargle)) (autoload (function flargle) "example" nil t)) (ignore (let* (... ... ... ...) (add-to-list ... ...) (define-key ... key ...))))
```
The issue here is the `(unless (fboundp (quote flargle)) (autoload (function flargle) "example" nil t))` which I believe shouldn't be there.

Of course the extra autoload doesn't actually run due to the `fboundp` check, but I don't see why the duplicated block is there at all.

This doesn't seem to happen when both forms have `:defer t` in them, which contradicts the documentation of `use-package-always-defer`.

npostavs:
fixes #341.
```
Don't pass a constant as the state

for use-package-process-keywords, because the function may modify the
list object.  Modifying a quoted constant can lead to unexpected side
effects (e.g. values from previous use-package forms end up in
subsequent ones).
```

---

UPDATE: added a rebase of #253, and a fix for the `declare-function` expression.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -1167,7 +1167,7 @@ this file.  Usage:
       (let ((body
              (macroexp-progn
               (use-package-process-keywords name args*
-                (and use-package-always-defer '(:deferred t))))))
+                (and use-package-always-defer (list :deferred t))))))
         (if use-package-debug
             (display-buffer
              (save-current-buffer
--- a/use-package.el
+++ b/use-package.el
@@ -46,7 +46,7 @@
 (eval-when-compile (require 'cl))
 (eval-when-compile (require 'regexp-opt))
 
-(declare-function package-installed-p 'package)
+(declare-function package-installed-p "package")
 
 (defgroup use-package nil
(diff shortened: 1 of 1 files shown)


npostavs:
Ping. I've rebased this and added 2 other minor fixes to this, please merge.

PR #349, opened 2016-05-26
Handle :mode ("foo") better

### LINKED ISSUE #348: Putting parens around a single string in :mode breaks forward-word globally
This works as expected:

``` elisp
(use-package scss-mode
  :mode ("\\.scss\\'" "\\.css\\'"))
```

This works as expected:

``` elisp
(use-package scss-mode
  :mode "\\.scss\\'")
```

This causes global errors when doing any operation that relies on word boundaries (`forward-word`, `kill-word`, etc):

``` elisp
(use-package scss-mode
  :mode ("\\.scss\\'"))
```

After evaluating that form, every time I do a kill-word, I get this:

`kill-word: Symbol's function definition is void: nil`

I'm not extremely familiar with elisp, but I don't see why the last one is invalid, and it took forever to track this down.

### LINKED ISSUE #399: All unit tests fail
Not sure what the unit tests are supposed to test, but right now they all fail:

```none
$ emacs -Q -batch -L . -l use-package-tests.el -f ert-run-tests-batch-and-exit
Running 2 tests (2016-10-31 12:12:23+0100)
Test use-package-mplist-get backtrace:
  (let ((fn-0 (function equal)) (args-1 (list (use-package-mplist-get 
  (lambda (test) (let ((fn-0 (function equal)) (args-1 (list (use-pack
  mapc((lambda (test) (let ((fn-0 (function equal)) (args-1 (list (use
  (let ((mplist (quote (:foo bar baz bal :blob plap plup :blam))) (tes
  (lambda nil (let ((mplist (quote (:foo bar baz bal :blob plap plup :
  ert--run-test-internal([cl-struct-ert--test-execution-info [cl-struc
  ert-run-test([cl-struct-ert-test use-package-mplist-get nil (lambda 
  ert-run-or-rerun-test([cl-struct-ert--stats t [[cl-struct-ert-test u
  ert-run-tests(t #[385 "\306\307\"\203G\211\211G\310U\203\211@\20
  ert-run-tests-batch(nil)
  ert-run-tests-batch-and-exit()
  command-line-1(("-L" "." "-l" "use-package-tests.el" "-f" "ert-run-t
  command-line()
  normal-top-level()
Test use-package-mplist-get condition:
    (void-function use-package-mplist-get)
   FAILED  1/2  use-package-mplist-get
Test use-package-mplist-keys backtrace:
  (let ((fn-5 (function equal)) (args-6 (list (use-package-mplist-keys
  (lambda nil (let ((fn-5 (function equal)) (args-6 (list (use-package
  ert--run-test-internal([cl-struct-ert--test-execution-info [cl-struc
  ert-run-test([cl-struct-ert-test use-package-mplist-keys nil (lambda
  ert-run-or-rerun-test([cl-struct-ert--stats t [[cl-struct-ert-test u
  ert-run-tests(t #[385 "\306\307\"\203G\211\211G\310U\203\211@\20
  ert-run-tests-batch(nil)
  ert-run-tests-batch-and-exit()
  command-line-1(("-L" "." "-l" "use-package-tests.el" "-f" "ert-run-t
  command-line()
  normal-top-level()
Test use-package-mplist-keys condition:
    (void-function use-package-mplist-keys)
   FAILED  2/2  use-package-mplist-keys

Ran 2 tests, 0 results as expected, 2 unexpected (2016-10-31 12:12:23+0100)

2 unexpected results:
   FAILED  use-package-mplist-get
   FAILED  use-package-mplist-keys
```

npostavs:
Fixes #348, fixes #399. Fixes `use-package-tests.el`

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -712,6 +712,10 @@ If RECURSED is non-nil, recurse into sublists."
 (defun use-package-normalize-binder (name keyword args)
   (use-package-as-one (symbol-name keyword) args
     (lambda (label arg)
+      (unless (consp arg)
+        (use-package-error
+         (concat label " a (<string or vector> . <symbol or string>)"
+                 " or list of these")))
       (use-package-normalize-pairs (lambda (k) (or (stringp k) (vectorp k)))
                                    (lambda (b) (or (symbolp b) (stringp b)))
                                    name label arg))))
--- a/use-package.el
+++ b/use-package.el
@@ -681,47 +681,40 @@ manually updated package."
 ;; :bind, :bind*
 ;;
 
-(defsubst use-package-is-sympair (x &optional allow-vector)
-  "Return t if X has the type (STRING . SYMBOL)."
+(defsubst use-package-is-pair (x car-pred cdr-pred)
+  "Return non-nil if X is a cons satisfying the given predicates.
+CAR-PRED and CDR-PRED are applied to X's `car' and `cdr',
+respectively."
   (and (consp x)
-       (or (stringp (car x))
-           (and allow-vector (vectorp (car x))))
-       (symbolp (cdr x))))
-
-(defsubst use-package-is-string-pair (x)
-  "Return t if X has the type (STRING . STRING)."
-  (and (consp x)
-       (stringp (car x))
-       (stringp (cdr x))))
+       (funcall car-pred (car x))
+       (funcall cdr-pred (cdr x))))
 
 (defun use-package-normalize-pairs
-    (name label arg &optional recursed allow-vector allow-string-cdrs)
-  "Normalize a list of string/symbol pairs.
-If RECURSED is non-nil, recurse into sublists.
-If ALLOW-VECTOR is non-nil, then the key to bind may specify a
-vector of keys, as accepted by `define-key'.
-If ALLOW-STRING-CDRS is non-nil, then the command name to bind to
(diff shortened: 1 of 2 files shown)


use-package.el
```diff
@@ -656,47 +656,40 @@ manually updated package."
 ;; :bind, :bind*
 ;;
 
-(defsubst use-package-is-sympair (x &optional allow-vector)
-  "Return t if X has the type (STRING . SYMBOL)."
+(defsubst use-package-is-pair (x car-pred cdr-pred)
```
jwiegley: This will affect all the other keywords that expect pairs too.  Do you have tests for all of them?

npostavs: I tested `use-package-normalize-mode` (`:mode`, `:interpreter`) and `use-package-normalize-binder` (`:bind`, `:bind*`). Perhaps `use-package-normalize-binder` could use a few more tests though.

npostavs:
I've added some more tests for `use-package-normalize-binder`. I found that my refactoring did actually make a change, but I believe it fixes another bug: `:bind ([keysym] . "string")` would actually bind keysym to `nil` (i.e., unbind it).  It now binds to "string" as expected.

Also, I noticed that `(use-package foopkg :bind "<some-key>")` was actually binding `<some-key>` to `foopkg`. I think this should be an error instead, so I made that additional change. I can drop the last commit if you disagree.

PR #350, opened 2016-06-05
Fix to ignore load error caused via :after

robario:
To load some related packages as order using :after.

```lisp
(use-package multi-web-mode
  :after web-mode
  :config
  (message "both web-mode and multi-web-mode are available"))
```

However, when web-mode is installed and multi-web-mode is not installed, this error occurred.
> (file-error "Cannot open load file" "no such file or directory" "multi-web-mode")


robario:
@jwiegley 
Awesome thanks!

PR #354, opened 2016-06-13
Add imenu support for use-package forms

justbur:
Also add require forms and group both under menu "Package".


jwiegley:
@justbur This is great! Would you, or have you already, sign a copyright assignment form with the FSF, so that I can include these changes into Emacs once use-package is included with it?

justbur:
Yes I've already signed them.

PR #355, opened 2016-06-14
Fix imenu support for older versions

justbur:
lisp-mode-symbol-regexp was not defined in Emacs 24.5.


justbur:
Sorry my mistake

PR #356, opened 2016-06-15
Improve imenu support

justbur:
Instead of using defvar for lisp-mode-symbol-regexp, wait until
lisp-mode is loaded and check for its existence to avoid making
use-package the place where this variable is declared.

By the way, sorry for all of the changes. I wasn't thinking carefully when I did this the first time.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -177,19 +177,17 @@ Must be set before loading use-package."
   :group 'use-package)
 
 (when use-package-enable-imenu-support
-  ;; Not defined in Emacs 24
-  (defvar lisp-mode-symbol-regexp
-    "\\(?:\\sw\\|\\s_\\|\\\\.\\)+")
-  (add-to-list
-   'lisp-imenu-generic-expression
-   (list "Package"
-         (purecopy (concat "^\\s-*("
-                           (eval-when-compile
-                             (regexp-opt
-                              '("use-package" "require")
-                              t))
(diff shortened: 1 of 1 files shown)


use-package.el
```diff
@@ -177,19 +177,18 @@ Must be set before loading use-package."
...
-         (purecopy (concat "^\\s-*("
-                           (eval-when-compile
-                             (regexp-opt
-                              '("use-package" "require")
-                              t))
-                           "\\s-+\\(" lisp-mode-symbol-regexp "\\)"))
-         2)))
+  (eval-after-load 'lisp-mode
+    `(let ((sym-regexp (or (and (boundp 'lisp-mode-symbol-regexp)
+                                lisp-mode-symbol-regexp)
```
npostavs: You can use `(or (boundp-and-true lisp-mode-symbol-regexp) ...)`

use-package.el
```diff
@@ -177,19 +177,18 @@ Must be set before loading use-package."
...
+                           "\\(?:\\sw\\|\\s_\\|\\\\.\\)+")))
+       (add-to-list
+        'lisp-imenu-generic-expression
+        (list "Packages"
+              (purecopy (concat "^\\s-*("
```
npostavs: `purecopy` is a no-op outside of preloaded files

justbur:
Thanks @npostavs. I incorporated your changes.

Issue #329, opened 2016-03-08
Feature Request: Function to find use-package declaration


justbur:
I think it would be really nice if there were a way to jump to the location of a use-package declaration for a given package. Not sure how difficult this is. Perhaps a list could record a marker position for each feature that is loaded with use-package? Thanks

justbur:
Or if there were just a way to generate an alist of packages with markers that could be fed into a completing read function that would be great.

jwiegley:
I could easily tag the package's symbol name with metadata identifying the file the use-package form came from.  I'm not so sure about identifying its file and line number.  I like this idea, though.

justbur:
With the file, a simple regexp search should be enough. Having the marker
would be cool though.

waymondo:
FWIW I have a personal solution for this in my `.emacs`. I have an `imenu` expression setup so I can quickly jump to any specific `use-package` declaration:

``` elisp
(defun use-package-imenu-elisp-expression ()
    (add-to-list 'imenu-generic-expression
                 '("packages" "^\\s-*(\\(use-package\\)\\s-+\\(\\(\\sw\\|\\s_\\)+\\)" 2)))
  (add-hook 'emacs-lisp-mode-hook #'use-package-imenu-elisp-expression)
```

I also have `ido-ubiquitous` and `imenu-anywhere` installed so I get a nice completing read from any `emacs-lisp-mode` file, but built-in `imenu` support might be cool since it is a native package.

justbur:
imenu would be nice, but this feature would be much more useful to me if it
could open the file the declaration is in. I don't have them all in one
file.

PR #359, opened 2016-06-23
Add function use-package-jump-to-package-form

justbur:
This is an attempt at resolving #329. The new interactive function
use-package-jump-to-package-form will prompt with a completing read of
all known packages. After selecting a package, use-package-find-require
searches load-history to see where the package was required and then I
attempt to find the correct use-package form using
use-package-form-regexp.

It will fail if the use-package form you are looking for did not
actually load the package. For example it could be something that is a
dependency of a library that was already loaded. In some sense this is a
feature because it is helpful to know that the library was already
loaded when your use-package form was encountered. It will also fail if
your use-package declaration doesn't match the regexp used, but this is
easily adjusted.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -191,6 +191,47 @@ Must be set before loading use-package."
                            "\\s-+\\(" lisp-mode-symbol-regexp "\\)"))
          2)))
 
+(defvar use-package-form-regexp "^\\s-*(\\s-*use-package\\s-+\\_<%s\\_>"
+  "Regexp used in `use-package-jump-to-package-form' to find use
+package forms in user files.")
+
+(defun use-package--find-require (package)
+  "Find file that required PACKAGE by searching
+`load-history'. Returns an absolute file path or nil if none is
+found."
+  (catch 'suspect
+    (dolist (filespec load-history)
+      (dolist (entry (cdr filespec))
+        (when (equal entry (cons 'require package))
+          (throw 'suspect (car filespec)))))))
+
+(defun use-package-jump-to-package-form (package)
+  "Attempt to find and jump to the `use-package' form that loaded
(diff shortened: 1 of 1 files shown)


use-package.el
```diff
@@ -191,6 +191,47 @@ Must be set before loading use-package."
...
+      (setq file (if (string= (file-name-extension requiring-file) "elc")
+                     (concat (file-name-sans-extension requiring-file) ".el")
+                   requiring-file))
+      (when (file-exists-p file)
+        (with-temp-buffer
+          (insert-file-contents file)
+          (goto-char (point-min))
+          (setq location
+                (re-search-forward (format use-package-form-regexp package) nil t)))
+        (find-file-other-window file)
```
npostavs: Why are you using a temp buffer, if you are opening the file for real regardless? Wouldn't it make more sense to do the search in the file's buffer?

justbur: That's a good question. I initially was going to throw an error if the form couldn't be found (hence the temp buffer), but then I changed my mind and thought that maybe it would be useful to show the file anyway (which makes the temp buffer pointless). 

What is your opinion? Would you expect to jump to the file even if the use-package form does not appear to be there?

npostavs: > Would you expect to jump to the file even if the use-package form does not appear to be there?

Yes (though I don't feel strongly about it). I think I would probably want to inspect the file to see why it wasn't found.

justbur: Me too. I'll clean this up to get rid of the temp buffer

jwiegley:
Hi @justbur, do you have a copyright assignment on file with the FSF?  I would need one to accept this pull request.  If you are willing to sign and submit one, please contact assign@gnu.org.

justbur:
Hi, I do have one on file already

PR #367, opened 2016-07-20
Update use-package.el

ketbra:
`:pin` does not work with `:ensure`, because it doesn't add the package to package-pinned-packages until after reading the package archive contents.  This change causes the package archive contents to be reread if the package is pinned and `:ensure` is being used.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -500,6 +500,8 @@ manually updated package."
 (defun use-package-ensure-elpa (package &optional no-refresh)
   (if (package-installed-p package)
       t
+    (if (and (not no-refresh) (assoc package package-pinned-packages))
+        (package-read-all-archive-contents))
     (if (or (assoc package package-archive-contents) no-refresh)
         (package-install package)
       (progn


npostavs:
I haven't tested or anything, but could you use spaces instead of tabs for indenting? Since lisp code doesn't get indented in multiples of tab width, it looks bad when moving to an environment with a different tab width (like Github, which uses 4-space tabs).

ketbra:
Woops.  Fixed!  I guess that's what I get for developing and testing the setup with `emacs -Q` .  Thanks for taking a look.

npostavs:
Yeah, looks good to me. I guess this shouldn't add too much overhead unless you have a lot of pinned packages and they're all being installed in which case you already expect to wait a long time for the installation anyway.

PS you may as well squash and `git push -f`. I'm usually in favour of showing intermediate commits, but in cases of trivial whitespace fixes, it's not so useful.

> I guess that's what I get for developing and testing the setup with `emacs -Q` .

The defaults should probably be changed. Though I hesitate to bring it up on emacs-devel in case it sparks another pointless tabs vs spaces debate.

ketbra:
> PS you may as well squash and git push -f. I'm usually in favour of showing intermediate commits, but in cases of trivial whitespace fixes, it's not so useful.

Sure thing.  I squashed the commits!  

> The defaults should probably be changed. Though I hesitate to bring it up on emacs-devel in case it sparks another pointless tabs vs spaces debate.

Ha, yeah, there are no shortage of opinions when it comes to better defaults vs backwards compatibility.

PR #369, opened 2016-07-25
Make :ensure working for package that are built-in and also on (m)elpa.

thierryvolpiatto:
Typically org package.

* use-package.el (use-package-ensure-elpa): Do it.
package-installed-p is of no help, use another test.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -486,6 +486,7 @@ manually updated package."
 ;; :ensure
 ;;
 (defvar package-archive-contents)
+(defvar package-alist)
 (declare-function package-desc-status "package")
 (declare-function package-read-all-archive-contents "package")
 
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
This fix:

`(use-package org :ensure t)`

Note: I haven't tested with a recent emacs, perhaps it works, also my version of package.el is slighly different than the emacs-24.5 one, so perhaps it is working for you without this patch, in this case just ignore the PR.

BTW this raise the questions:
Why do we try to handle built-in packages instead of just ignoring them as they are part of emacs?
Why instead not moving them to elpa and remove them from core emacs?

thierryvolpiatto:
Still one problem, elpa/melpa are called even when packages are installed, so don't merge this PR until I have (or someone) fixed this.
Thanks.

thierryvolpiatto:
Now fixed.

npostavs:
> perhaps it is working for you without this patch,

I think you forgot to mention what you mean by "working" and "not working".

thierryvolpiatto:
Noam Postavsky <notifications@github.com> writes:

>     perhaps it is working for you without this patch,
>    
> I think you forgot to mention what you mean by "working" and "not working".

If you read the full post no.

What is not working is (use-package org :ensure t).

Is it working for you ?

Note: What is expected is the installation of org from elpa when
starting emacs.

-- 
Thierry

npostavs:
> Note: What is expected is the installation of org from elpa when starting emacs.

Ah, that's what was missing. I suggest you explain also in the commit message how you are changing the behaviour. And update the title of this issue, so people might have a clue what it's about.

It's not clear to me that `:ensure` *should* ignore builtin versions, but I will leave that for others to comment on.

thierryvolpiatto:
There is two things changed in this patch:

1) Check if package is installed by just looking in `package-alist` if it
   is present instead of using `package-installed-p` which is falling back
   to `package-built-in-p` when failing on first check.
   This allows to switch to the else part of if clause.

2) When switching to else part the second problem we hit is the call to
   `package-install` which use as argument the name of the package (org
   in our example) and it fail to install because the package (org) is
   already installed (builtin) so in the proposed patch I use as
   argument the full description of the package (the one from elpa this
   time).

With this explanation, does ":ensure should ignore builtin versions"
makes more sense ?

Actually with this patch the following behaviors are observed:

- org package is installed at emacs startup when `:ensure t` is given
in the `use-package` form.
- when starting emacs with all user packages installed `use-package` is
not calling remote elpa.
- the packages present in package-pinned-packages are installed from the
right archives.

Let me know if something isn't correct.

-- 
Thierry

thierryvolpiatto:
I have squashed all commits and made a new PR #370 for this.

PR #370, opened 2016-07-28
Allow :ensure installing built-in package distributed also on (m)elpa.

thierryvolpiatto:
This commit fix (use-package org :ensure t)
org being a built-in package but also a package distributed on elpa.
When a user use :ensure on such a package (already installed) we assume
it is the elpa version he wants to use.

* use-package.el (use-package-ensure-elpa):
Don't assume a package is installed because it is built-in.
Use the description of the package to ensure we don't try to install
the built-in one.

Add also some declarations to shutup byte compiler.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -486,6 +486,10 @@ manually updated package."
 ;; :ensure
 ;;
 (defvar package-archive-contents)
+(defvar package-alist)
+(declare-function package-desc-status "package")
+(declare-function package-read-all-archive-contents "package")
+
 (defun use-package-normalize/:ensure (name keyword args)
   (if (null args)
       t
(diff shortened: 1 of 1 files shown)


tsdh:
@thierryvolpiatto I guess there's the assumption that the ELPA package is the newer version of the built-in package.  I just want to mention that this does not need to be true.  One example is `cl-lib`.  The most up-to-date version is in emacs proper (version 1.0 in master), and the ELPA package (version 0.5) is just for older emacsen.

So basically you might want to configure different things:

1. Take the newest version no matter if built-in or ELPA
2. Take the built-in version unless there's none
3. Take the ELPA version even though the built-in package's version is newer

IMHO, the current semantics of just sticking to the built-in package version if there is one is sensible for `:ensure t`.  But maybe something like `:ensure 'newest` and `:ensure 'package` could be supported to cater for the options 1 and 3 above.

thierryvolpiatto:
Tassilo Horn <notifications@github.com> writes:

> @thierryvolpiatto I guess there's the assumption that the ELPA package
> is the newer version of the built-in package. I just want to mention
> that this does not need to be true. One example is cl-lib. The most
> up-to-date version is in emacs proper (version 1.0 in master), and the
> ELPA package (version 0.5) is just for older emacsen.

Exactly, it is why I think that when one add :ensure t, it is that he
wants the elpa version and not the built-in version, no matter if elpa
version is newer or not.

> So basically you might want to configure different things:
>
>  1. Take the newest version no matter if built-in or ELPA
>  2. Take the built-in version unless there's none
>  3. Take the ELPA version even though the built-in package's version is newer
>
> IMHO, the current semantics of just sticking to the built-in package
> version if there is one is sensible for :ensure t. But maybe something
> like :ensure 'newest and :ensure 'package could be supported to cater
> for the options 1 and 3 above.

Perhaps, but for now I think that :ensure should always ensure the
version available on (m)elpa.

As a side note, I think that built-in packages have nothing to do in
package-alist unless they are removed from emacs and moved to elpa, but
that's another story...

--
Thierry

tsdh:
Well, IMO :ensure should ensure that a package will be there, and the simplest way to meet that requirement for built-in packages is to do nothing.

I can see that sometimes you always want to use the ELPA version of a package, but I think that should be realized with a new option or special value of :ensure.

PR #376, opened 2016-08-15
Ensure package-pinned-packages is bound before referencing it.

### LINKED ISSUE #375: Lisp error: (void-variable package-pinned-packages) on Emacs 24.3
I believe this was introduced in #367.

I was able to work around it by adding the following to my init file:

```elisp
(unless (boundp 'package-pinned-packages)
  (defvar package-pinned-packages '()))
```

# Environment

Ubuntu 14.04.4
Emacs 24.3.1
use-package-20160722.1123

# Minimal init.el to repro

```elisp
(require 'package)
(add-to-list 'package-archives '("melpa" . "http://melpa.org/packages/"))
(package-initialize)
(unless (package-installed-p 'use-package)
   (package-refresh-contents)
   (package-install 'use-package))
(require 'use-package)
(use-package easy-kill :ensure t)
```

# Backtrace

```
Debugger entered--Lisp error: (void-variable package-pinned-packages)
  use-package-ensure-elpa(easy-kill)
  (progn (require (quote package)) (use-package-ensure-elpa (quote easy-kill)))
  (progn (progn (require (quote package)) (use-package-ensure-elpa (quote easy-kill))) (if (not (require (quote easy-kill) nil (quote t))) $
  (use-package easy-kill :ensure t)
  eval((use-package easy-kill :ensure t) nil)
  eval-last-sexp-1(t)
  eval-last-sexp(t)
  eval-print-last-sexp()
  call-interactively(eval-print-last-sexp nil nil)
```

appleby:
Add a boundp guard to package-pinned-packages before referencing it in
use-package-ensure-elpa.

I'm just duplicating the boundp guard from use-package-pin-package. Let me know if you prefer another approach.

I've tested this on Ubuntu 14.04 and verified it no longer throws an error, but haven't tested much else. Specifically, I don't really use the :pin feature myself, and haven't actually tested to ensure that package pinning works as expected.

Fixes #375

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -500,7 +500,8 @@ manually updated package."
 (defun use-package-ensure-elpa (package &optional no-refresh)
   (if (package-installed-p package)
       t
-    (if (and (not no-refresh) (assoc package package-pinned-packages))
+    (if (and (not no-refresh)
+             (assoc package (bound-and-true-p package-pinned-packages)))
         (package-read-all-archive-contents))
     (if (or (assoc package package-archive-contents) no-refresh)
         (package-install package)


npostavs:
> Specifically, I don't really use the :pin feature myself, and haven't actually tested to ensure that package pinning works as expected.

It wouldn't work on 24.3, because the package.el there doesn't implement it. Not throwing an error is about as good as can be expected.

> I'm just duplicating the boundp guard from use-package-pin-package. Let me know if you prefer another approach.

Personally, I would go with just wrapping the reference in `bound-and-true-p`, but that's just a minor style thing.

appleby:
I've pushed an updated commit that uses bound-and-true-p.  Based on your comments in #367, I've squashed the commit into my previous change. It seems like github is automatically picking up the new commit in this pull request, but let me know if I need to do something else to update this thread.

npostavs:
Looks good to me.

PR #418, opened 2016-12-16
Increase :preface priority

basil-conto:
The `README` describes the `:preface` keyword as occurring before everything but `:disabled`, yet the former is currently preceded by seven other keywords in addition to the latter.

Most importantly, `:preface` occurs after `:ensure` and the three conditional keywords, thus reducing its use for establishing definitions.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -126,6 +126,7 @@ the user specified."
 
 (defcustom use-package-keywords
   '(:disabled
+    :preface
     :pin
     :ensure
     :if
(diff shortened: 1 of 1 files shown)


jwiegley:
Are you sure you want it before `:if`?

basil-conto:
Personally, I am impartial to the ordering of `:preface` with respect to the three conditional keywords. My understanding of its semantics, however, is based solely on its description in the `README`, so there is a good chance I could be missing something.

As the PR mentions, `:preface` is described in the `README` as "occurring before everything except `:disabled`" and its purpose as being to "allow you to define code that can be used in an `:if` test."

I would, therefore, be grateful for clarification on the intended purpose of `:preface` and the drawbacks of it occurring before `:if`.

PR #419, opened 2016-12-17
Delay decision whether to use eval-after-load until run-time

### LINKED ISSUE #378: Use eval-after-load for :bind (:map ...)
I tried adding this to my init file:

``` lisp
(use-package magit-mode
  :defer t
  :bind (:map magit-mode-map
              ("f" . magit-pull-and-fetch-popup)
              ("F" . nil)))
```

which expands to:

``` lisp
(progn
  (unless
      (fboundp 'magit-pull-and-fetch-popup)
    (autoload
      (function magit-pull-and-fetch-popup)
      "magit-mode" nil t))
  (ignore
   (progn
     (bind-key "f" 'magit-pull-and-fetch-popup magit-mode-map nil)
     (bind-key "F" 'nil magit-mode-map nil))))
```

and results in this warning:

```
Symbol's value as variable is void: magit-mode-map
```

Have a nice day!

tarsius:
```
Just because a keymap variable is bound at macro-expansion-time
doesn't mean that it must be bound at run-time too.

Change `bind-keys-form', which is used by `bind-keys' and other
macros, to return a form which delays the decision on whether to
wrap the binding forms with `eval-after-load' until run-time.
```

Fixes #378.

### DIFF
--- a/bind-key.el
+++ b/bind-key.el
@@ -241,9 +241,9 @@ function symbol (unquoted)."
       (cl-flet
           ((wrap (map bindings)
                  (if (and map pkg (not (eq map 'global-map)))
-                     (if (boundp map)
-                         bindings
-                       `((eval-after-load
+                     `((if (boundp ',map)
+                           (progn ,@bindings)
+                         (eval-after-load
                              ,(if (symbolp pkg) `',pkg pkg)
                            '(progn ,@bindings))))
                    bindings)))


tarsius:
Updated to take https://github.com/jwiegley/use-package/issues/378#issuecomment-267766488 into account.

PR #426, opened 2017-01-16
Add use-package-defaults

raxod502:
Previously, the :config, :ensure, and :pin keywords had default
values (dependent on the values of the use-package-always-ensure and
use-package-always-pin). This change allows the user to customize the
default values used for those keywords, and add default values for
their own keywords in a non-hacky way.

This functionality would be useful for (as an example) the
quelpa-use-package package, which needs to use an advice to override
the functionality of :ensure. The same problem prevents adding a
use-package-always-quelpa variable in any reasonable way, without a
way to customize the default values of keywords.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -192,6 +192,20 @@ The default value uses package.el to install the package."
                  (function :tag "Custom"))
   :group 'use-package)
 
+(defcustom use-package-defaults
+  '((:config '(t) t)
+    (:ensure use-package-always-ensure use-package-always-ensure)
+    (:pin use-package-always-pin use-package-always-pin))
+  "Alist of default values for `use-package' keywords.
+Each entry in the alist is a list of three elements. The first
+element is the `use-package' keyword and the second is a form
+that can be evaluated to get the default value. The third element
+is a form that can be evaluated to determine whether or not to
+assign a default value; if it evaluates to nil, then the default
+value is not assigned even if the keyword is not present in the
+`use-package' form."
+  :type '(repeat (list symbol sexp sexp)))
+
 (when use-package-enable-imenu-support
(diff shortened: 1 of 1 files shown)


raxod502:
Rebased to fix conflicts.

raxod502:
Any update on this?

jwiegley:
@raxod502 Have you submitted a copyright assignment form to the FSF for your work on Emacs?  That's all I would need before merging, since this code is intended to be shipped with Emacs in the near future.  And thank you very much for the change and the idea!

raxod502:
I have emailed assign@gnu.org.

jwiegley:
Great!  Let me know once they've confirmed, and I'll merge.

raxod502:
@jwiegley They have confirmed my copyright assignment. I am disappointed there seems to be no straightforward way for me to demonstrate that, but let me know if there is anything I can do.

jwiegley:
@raxod502 Hi, as a GNU project maintainer I can access the master list directly, and I see your assignment on file. Thank you!

jwiegley:
@raxod502 I've received reports in two other bugs now that some keywords have suddenly become ordering dependent: i.e., unless A is used before B, it does not take effect.  They claim it's due to this change, so I may need to revert it for the time being until we figure out what's wrong.

jwiegley:
Reverted the merge in fe736725dee694e25a944dd6368b5774fa347867.

jwiegley:
Problem was seen in #429 and #428.

raxod502:
That's very concerning! I didn't understand why the code was set up the way it was, so I tried to replicate the exact logic. Clearly I failed, let me take a look at it again.

Issue #429, opened 2017-02-13
`:diminish` stopped working


joostkremers:
I've been using the `:diminish` keyword extensively to get rid of minor-mode indicators, which has been working fine, but since the last update, `:diminish` seems to have issues. If I evaluate a `use-package` declaration with `C-M-x`, it works, but when I start Emacs, `:diminish` seems to be ignored.

I haven't changed anything in my Emacs config the past few days, other than updating my packages. Could it be that the latest changes in `use-package` caused `:diminish` to be ignored, or should I be looking for something in my setup? In the latter case, some hints where I could start debugging this would be very much appreciated.

drot:
Can confirm this and it only happens if you use it in conjunction with `:ensure`, then you have to have `:diminish` as the last keyword in order for it to work.

gongzhitaao:
For me it happens whatsoever unless you put it last.  For example

The following code **works** for me:

```emacs-lisp
(use-package which-key
  :config (which-key-mode)
  :diminish which-key-mode)
```

But the following code **does not** work for me

```emacs-lisp
(use-package which-key
  :diminish which-key-mode
  :config (which-key-mode))
```

jwiegley:
Was this caused by 9a4e5cbe3601361188a71bc37284af6e4adb68cc?

dakrone:
> Was this caused by 9a4e5cb?

That is what it looks like. I updated `use-package` this morning and am running into this same problem.

jwiegley:
I've reverted that commit for the time being.

jwiegley:
Should be fixed now.

joostkremers:
:thumbsup:

Issue #428, opened 2017-02-13
:demand keyword needs to be put AFTER :bind


lurdan:
In my case, when I put `:demand t` before `:bind`, expanded macro doesn't include `require` part (which `use-package-debug` shows).

Though I may be overlook usage, are there any requirements about keywords?

anler:
Don't know exactly why but I had this:

```elisp
(use-package prelude
  :load-path "~/.emacs.d/core"
  :demand t
  :bind* (...)
  ...)
```

and now had to move the `demand` after the `bind*` in order to be able to *demand* the package

greenonion:
This has been introduced in 9a4e5cbe3601361188a71bc37284af6e4adb68cc.

jwiegley:
I've reverted this commit for the time being.

jwiegley:
Should be fixed now.

anler:
@jwiegley thanks!

PR #430, opened 2017-02-13
Fix use-package-defaults

raxod502:
This pull request is a continuation from https://github.com/jwiegley/use-package/pull/426#issuecomment-279479609 in response to #428 and #429, which it should fix.

The problem was indeed that I didn't quite replicate the exact logic of the original version. In the original version, for each keyword, `use-package-plist-maybe-put` was only called if the relevant predicate returned true, whereas `use-package-sort-keywords` was called unconditionally. In the new version, `use-package-sort-keywords` was also called only when the relevant predicate returned true.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -192,6 +192,20 @@ The default value uses package.el to install the package."
                  (function :tag "Custom"))
   :group 'use-package)
 
+(defcustom use-package-defaults
+  '((:config '(t) t)
+    (:ensure use-package-always-ensure use-package-always-ensure)
+    (:pin use-package-always-pin use-package-always-pin))
+  "Alist of default values for `use-package' keywords.
+Each entry in the alist is a list of three elements. The first
+element is the `use-package' keyword and the second is a form
+that can be evaluated to get the default value. The third element
+is a form that can be evaluated to determine whether or not to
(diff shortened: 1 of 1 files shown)


raxod502:
I'd like to request @jwiegley to take a look at the code since I had to make an educated guess about how to generalize from the n = 3 case previously implemented. (The guess I made was that `use-package-normalize-plist` is called unconditionally first, and then for each keyword with a default value, `use-package-plist-maybe-put` is called conditionally on the relevant predicate (e.g. `use-package-always-ensure`) and `use-package-sort-keywords` is called unconditionally.)

@lurdan @anler @greenonion @joostkremers @drot @gongzhitaao @dakrone
Does this patch fix the issues you are all encountering? I reproduced the issue described by @gongzhitaao and confirmed that this patch fixes that particular issue, but I can't be sure about the others yet.

I am very sorry for the inconvenience.

dakrone:
@raxod502 preliminary testing seems to indicate this fixes the issue(s) for me, I've only tried it for a few minutes, but I'll keep testing with it; thanks!

anler:
I confirm that it works for me, thanks a lot @raxod502 !

Issue #409, opened 2016-11-21
Defer installation of a package?


raxod502:
I have no idea what the most elegant semantics for this would be, or if it is even remotely feasible, but I think it would be really cool to be able to lazy-*install* packages. That is, `markdown-mode` would not be installed from MELPA until you opened a `.md` file, and so on.

Here's some example usage:

```
(use-package cider
  :defer-install ("\\.clj\\'"
                  cider-jack-in)
  :ensure t)

(global-set-key (kbd "C-c M-j") 'cider-jack-in)
```

This would ignore everything except for `:defer-install` in the `use-package` form. But, when a `.clj` file is opened, the `use-package` form would be evaluated (if it had not been already), causing the package to be installed. The above declaration would also generate a shim for `cider-jack-in` that would just evaluate the `use-package` form (if it had not been already) and then call `cider-jack-in` (which would presumably be defined by the loaded package) for real.

This feature would allow an Emacs configuration to include packages for many different languages without taking absolutely forever to download all of them on first launch.

raxod502:
Well, I guess this feature has been added now :D

PR #433, opened 2017-03-09
[RFC] Support for deferred installation

raxod502:
First cut at :defer-install keyword

    This new keyword, if provided along with a non-nil value, causes the action
    of :ensure to be deferred until "necessary". Package installation can be
    triggered by the user calling the new interactive function
    `use-package-install-deferred-package', or by the feature declared by the
    `use-package' form being required. This latter behavior seems to be the
    simplest way to make sure that package installation actually takes place
    when it needs to, but it requires that an advice be added to `require',
    which may be considered overly intrusive. (Also, it's generally considered
    bad practice for functions in Emacs to put advice on other functions in
    Emacs.) Thus it may make sense to add an option or function to explicitly
    enable this behavior, if there does not turn out to be a better way to
    accomplish deferred installation.

    Documentation has not been updated to reflect :defer-install yet.

Considering how to address #409. I think that this is not the right
approach to take (esp. because autoloads are handled in the C code and
therefore do not respect the advice on `require`), but I thought I
would post my progress so far. I think I have an idea about how to
proceed.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -185,21 +185,43 @@ Must be set before loading use-package."
 
 (defcustom use-package-ensure-function 'use-package-ensure-elpa
   "Function that ensures a package is installed.
-This function is called with three arguments: the name of the
+This function is called with four arguments: the name of the
 package declared in the `use-package' form; the argument passed
-to `:ensure'; and the current `state' plist created by previous
-handlers. Note that this function is called whenever `:ensure' is
-provided, even if it is nil. It is up to the function to decide
-on the semantics of the various values for `:ensure'.
+to `:ensure'; the current `state' plist created by previous
+handlers; and a keyword indicating the context in which the
+installation is occurring.
 
-The default value uses package.el to install the package."
+Note that this function is called whenever `:ensure' is provided,
+even if it is nil. It is up to the function to decide on the
+semantics of the various values for `:ensure'.
+
+This function should return non-nil if the package is installed.
+
+The default value uses package.el to install the package.
+
+Possible values for the context keyword are:
+
+:byte-compile - package installed during byte-compilation
+:ensure - package installed normally by :ensure
+:autoload - deferred installation triggered by an autoloaded
+            function
+:after - deferred installation triggered by the loading of a
+         feature listed in the :after declaration
+:config - deferred installation was specified at the same time
+          as :demand, so the installation was triggered
+          immediately
+:unknown - context not provided
+
+Note that third-party code can provide other values for the
+context keyword by calling `use-package-install-deferred-package'
+with the appropriate value."
   :type '(choice (const :tag "package.el" use-package-ensure-elpa)
                  (function :tag "Custom"))
   :group 'use-package)
 
 (defcustom use-package-pre-ensure-function 'ignore
   "Function that is called upon installation deferral.
-It is called with the same arguments as
+It is called immediately with the same arguments as
 `use-package-ensure-function', but only if installation has been
 deferred. It is intended for package managers other than
 package.el which might want to activate the autoloads of a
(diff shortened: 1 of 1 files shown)


raxod502:
It works! Here is a short demo: https://youtu.be/az5C1nd3fh0

Usage looks like this:

```
(use-package swift-mode
  :defer-install t
  :mode "\\.swift\\'")
```

In theory, specifying `:defer-install t` should "just work", in that all entry points to loading the package normally created by `use-package` will now prompt the user to install the package before requiring the feature.

I'll be testing this in my own config. Comments and suggestions would be appreciated!

raxod502:
I've realized that the confirmation prompt needs to be moved out of the `use-package` logic and into to the package-manager-specific `:ensure` function, since otherwise there is no way to only prompt the user if the package is *not* installed.

raxod502:
It appears to be at least mostly working! I have integrated deferred installation with [my config](https://github.com/raxod502/radian/tree/develop/radian-emacs), and out of 52 packages only 25 are installed on initial Emacs launch.

raxod502:
Wow, how did I base this branch on a four-week-old commit? I've fixed the merge conflicts.

drot:
After this merge the `:after` keyword stopped working, if I have something like this:

```
(use-package info+
  :ensure t
  :after info)
```
info+ doesn't get loaded anymore.

raxod502:
That's not good—`:defer-install` should not affect behavior at all unless you explicitly include it in the `use-package` form. Let me see if I can reproduce the issue.

raxod502:
@drot Should be fixed now, it was a silly quoting error.

seagle0128:
Can you please add `use-package-always-defer-install`  as `use-package-always-defer` for this feature?

raxod502:
This would be trivial to add via `use-package-defaults`. One thing to note
though, is that autoloads are no longer automatically accessible when the
package is yet to be installed (because obviously if the source code isn't
on your hard drive, there's no way to tell what autoloaded commands the
package provides). Same for `auto-mode-alist`, etc. So you have to set
these by hand (using `:commands`, `:mode`, etc.) if you want to use
deferred installation.

blaenk:
Just found out about this through the commit log and I think it's pretty cool! Thanks for taking the initiative to implement it!

I was wondering if there's a way to have something like `use-package-always-ensure`, or some way to have behavior as "always assume `:ensure t` unless `:ensure` is explicitly present in a `use-package` form which specifies otherwise (e.g. `nil` or a package name), and beyond that assume `:defer-install t` so that the package is only installed if it's needed because e.g. it was `require`d, unless the `use-package` form explicitly has `:defer-install nil` or something."

So I guess said another way, assume `:defer-install t` unless it's overridden in the `use-package` declaration, and otherwise assume `:ensure t` in a similar manner (as already covered by `use-package-always-ensure`).

I was reading through the diff's elisp doc strings but couldn't figure out if this was already possible. If not, is it even possible? Or aside from that, is there a reason why that would be inadvisable?

raxod502:
@blaenk This is a trivial change so I made a pull request, see #443. One thing to note is that `:defer-install t` does nothing unless `:ensure t` is also given (or, of course, `use-package-always-ensure` is non-nil).

TrunovS:
@raxod502 It seems that change removed from master now. Is deferred install now possible in another way?

raxod502:
No, there's no way to achieve this behavior. If the feature were to be reimplemented, it should be done at the package-manager level, in my opinion. The ideal place would be `straight.el`, where there is already a mechanism for saving autoloads persistently (sort of) and it would be easy to extend the existing code to generate deferred-installation stubs for all entry points to the package, not just the ones mentioned in `use-package`: a better solution.

PR #438, opened 2017-03-19
Fix :after keyword

raxod502:
This should address [this issue](https://github.com/jwiegley/use-package/pull/433#issuecomment-287606553).


raxod502:
Argh! I hate when [`hub`](https://github.com/github/hub) gets the wrong branch. See #439.

Issue #442, opened 2017-03-26
Unable to start after byte-compiling with :defer-install


seagle0128:
GNU Emacs 25.1, macOS Sierra, lastest use-package

I got this error after adding :defer-install keywords and re-compiling el files.

Example
```elisp
(use-package counsel
  :defer-install t
  :diminish ivy-mode counsel-mode
  :bind (("C-s" . swiper)
         ("C-c u" . swiper-all))
```
Error:
```
Symbol's function definition is void: use-package-install-deferred-package
```

- if I use `(setq use-package-always-ensure nil)` and compile, it works.
- If I delete elc files, and restart, it works.

raxod502:
Ah, this probably slipped by me because I don't byte-compile my init files. Deferred installation will not work if you only load `use-package` at compile-time. The main problem is that, at some point, the deferred installation will have to actually happen, and `use-package` provides the code for it.

I think it might be possible to refactor so that this use case would work, though. We would need to:

1. change the interface for `use-package-ensure-function` so that it returns a form to be evaluated to install the package, instead of actually installing the package
2. put lambdas on the relevant hooks, instead of calls to `use-package-install-deferred-package`.

Alternatively, we could just autoload `use-package-install-deferred-package`, on the assumption that package installation being needed is not the norm. But that might not work, since `use-package-install-deferred-package` is called in a couple of places even when a package does not need to be installed.

Any thoughts?

seagle0128:
Thanks for your reply!
In my opinion, option 1 may work but I am not very sure. Option 2 is not acceptable.
Now I removed all .elc files and disabled to byte-compile my init file. it won't take much more time while starting.

raxod502:
A few comments:

* Why do you say option 2 is not acceptable?
* I don't think you have to disable byte-compilation. You just have to `(require 'use-package)` in your init-file, if you want to use deferred installation. Loading `use-package` will be pretty darn fast, I think (consider all the other features that are loaded by default).
* Regardless of what we do here, it will not affect usage of `use-package` without deferred installation (which will remain the same as before).

seagle0128:
Thanks, @raxod502.
I think option2 is not easy to use for the customers. The word "not acceptable" is not accurate. 
I disable byte-compilation for now, but I will try to use `(require 'use-package)`.

raxod502:
@seagle0128 Why would option 2 be hard to use? It would make everything work transparently, with no changes required in your init-file.

[...]

PR #443, opened 2017-03-26
Add use-package-always-defer-install

raxod502:
See https://github.com/jwiegley/use-package/pull/433#issuecomment-289317875

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -85,13 +85,18 @@ The check is performed by looking for the module using `locate-library'."
   :type 'boolean
   :group 'use-package)
 
+(defcustom use-package-always-defer-install nil
+  "If non-nil, assume `:defer-install t` unless `:defer-install nil` is given."
+  :type 'boolean
+  :group 'use-package)
+
 (defcustom use-package-always-ensure nil
   "Treat every package as though it had specified `:ensure SEXP`."
   :type 'sexp
   :group 'use-package)
 
 (defcustom use-package-always-pin nil
-  "Treat every package as though it had specified `:pin SYM."
+  "Treat every package as though it had specified `:pin SYM`."
   :type 'symbol
   :group 'use-package)
 
@@ -238,6 +243,9 @@ when the packages are actually requested."
 (defcustom use-package-defaults
   '((:config '(t) t)
     (:ensure use-package-always-ensure use-package-always-ensure)
+    (:defer-install
+     use-package-always-defer-install
+     use-package-always-defer-install)
     (:pin use-package-always-pin use-package-always-pin))
   "Alist of default values for `use-package' keywords.
 Each entry in the alist is a list of three elements. The first


blaenk:
Amazing! Thank you very much again for thinking up this idea and following through with implementing it!

raxod502:
@blaenk I'm glad you like it 😄. Ping me if any issues turn up with deferred installation, since it's a pretty crazy hack and I haven't done extensive testing.

blaenk:
So I just now got the chance to try this out, but I think I'm misunderstanding something.

So I have it set up to always `:ensure t` unless specified otherwise and same with `:defer-install t`. The thing is that there are some "utility" packages which I _do_ want to eager-install, such as general.el, s.el, dash.el, f.el, etc. I figured the way to opt-out of the assumed `:defer-install` was to specify `:defer-install nil`, but it doesn't seem to be working out. I even also explicitly specified `:ensure t` in case that was related, but it doesn't seem to help.

``` elisp
(setq
      use-package-always-ensure t
      use-package-always-defer-install t
      use-package-check-before-init t
      use-package-enable-imenu-support t)

;; ...

(use-package f :defer-install nil)
(use-package s :defer-install nil)

(use-package general :ensure t :defer-install nil)
```

After launching like this for the first time without an `elpa/` dir, a subsequent `(require 'general)` fails because the package wasn't installed and instead ended up being deferred. I notice that `use-package--deferred-packages` contains the packages I'm trying to have opt-out from `:defer-install`:

``` elisp
#s(hash-table size 65 test eql rehash-size 1.5 rehash-threshold 0.8 data
	      (benchmark-init
	       (t :defer-install :ensure :deferred t)
	       auto-compile
	       (t :defer-install :ensure)
	       dash
	       (t :defer-install :ensure)
	       f
	       (t :defer-install :ensure)
	       s
	       (t :defer-install :ensure)
	       general
	       (t :defer-install :ensure)))
```

I'm pretty sure I'm simply misunderstanding this. I would appreciate any clarification, thanks!

raxod502:
@blaenk I can reproduce the problem; you have found a bug. It appears to be unrelated to `use-package-always-defer-install`; the following does not install package `0xc` as it should, but the installation takes place correctly if `:defer-install nil` is removed:

```
(let ((repos-dir (concat user-emacs-directory "straight/repos/")))
  (load (concat repos-dir "straight.el/bootstrap.el")
        nil 'nomessage))
(straight-use-package 'use-package)

(use-package 0xc
  :ensure t
  :defer-install nil)
```

blaenk:
Thanks for confirming! Though I can't personally reproduce having it install by only keeping `:ensure t`:

``` elisp
(use-package general :ensure t)
```

raxod502:
See #449.

raxod502:
I should clarify the behavior, though, since it is currently undocumented.

Setting `use-package-always-ensure` to some non-nil value has the effect of providing the `:ensure` key with that value if and only if `:ensure` is not provided (with any value, nil or non-nil) in the `use-package` form. Same for `use-package-always-defer-install`.

If `:ensure` is nil, the package manager is not invoked under any circumstances. If `:defer-install` is non-nil, then the deferred installation routines will be activated *if* `:ensure` is also non-nil (otherwise nothing happens). If `:defer-install` is nil, then it has the same effect as being omitted.

In other words, your original understanding was entirely correct.

(I know all the gory details because I happened to implement `use-package-defaults` too! :D)

blaenk:
Thanks for that fix! I can confirm that it fixes the opt-out behavior.

Right now I'm trying to figure out how byte-compiling a file interacts with this feature. For example I've narrowed down the problem to this content in a file:

``` elisp
(eval-when-compile
  (require 'use-package))

(use-package company)

(provide 'conf/company)
```

This file is in a directory that I'm byte-compiling in my init.el with `byte-recompile-directory`. This however results in a failure to require `'company`:

error

I think this may be a misunderstanding of byte-compiling on my part, though it's not a problem I had before. Would you happen to have any ideas on what may be causing this?

raxod502:
Unfortunately, deferred installation is not yet compatible with loading `use-package` only at compile-time, see https://github.com/jwiegley/use-package/issues/442#issuecomment-290926047. You can fix the problem for now by using

```
(require 'use-package)
```

instead of

```
(eval-when-compile
  (require 'use-package))
```

but obviously that's not ideal.

blaenk:
Oh I don't mind about that, I had actually changed it to `eval-when-compile` because I was wondering if it was the cause of the problem, but yeah I'm getting that error with or without it. I also removed the `eval-when-compile` in my `init.el` which is the file that contains the `byte-recompile-directory` and I still get the error. I'll continue to attempt to narrow the issue.

raxod502:
Oh interesting. I think that this problem should be solvable without actually fixing #442 then. I'll have time to look at it more later. Let me know if you find anything out.

blaenk:
Sure no worries. I appreciate the help!

In case it helps, this narrowed down code still yields that error:

My init.el:

``` elisp
(require 'package)

;; packages
(add-to-list 'package-archives '("melpa" . "http://melpa.org/packages/") t)

(package-initialize)

(unless (package-installed-p 'use-package)
  (package-refresh-contents)
  (package-install 'use-package))

(setq load-prefer-newer t
      backup-by-copying t

      use-package-always-ensure t
      use-package-always-defer-install t
      use-package-check-before-init t
      use-package-enable-imenu-support t)

(require 'use-package)

(defsubst use-package-normalize-value (label arg)
  "Normalize a value."
  (cond ((null arg) nil)
        ((symbolp arg)
         `(symbol-value ',arg))
        ((functionp arg)
         `(funcall #',arg))
        (t arg)))

(require 'diminish)
(require 'bind-key)

(use-package benchmark-init
  :defer-install nil
  :defer t

  :init
  (defvar my-benchmarking-p
    (prog1
        (member "--benchmark" command-line-args)
      (setq command-line-args (delete "--benchmark" command-line-args)))
    "Whether emacs should run benchmark-init"))

(use-package auto-compile
  :defer-install nil

  :defines
  auto-compile-display-buffer

  :init
  (setq auto-compile-display-buffer nil)

  :config
  (auto-compile-on-load-mode)
  (auto-compile-on-save-mode))

(use-package dash
  :defer-install nil

  :config
  (dash-enable-font-lock))

(use-package f :defer-install nil)
(use-package s :defer-install nil)

(use-package general :defer-install nil)

(add-to-list 'load-path (expand-file-name "inits/" user-emacs-directory) t)

(byte-recompile-directory (expand-file-name "inits/" user-emacs-directory) 0)
```

.emacs.d/inits/conf/company.el

``` elisp
(require 'use-package)

(use-package company)

(provide 'conf/company)
```

raxod502:
With your code I get the expected behavior. (I.e. company is not installed or available by default, but after I install it and restart Emacs, it's loaded; either way, no error.) Perhaps you can try the good old `rm -rf ~/.emacs.d/elpa` trick.

I'm testing with 25.1; is your Emacs older?

blaenk:
> [...] but after I install it [...]

Would you mind clarifying? I should have stated my expectations, in case there's a disconnect in my expected behavior.

What I was intending to have happen is, aside from the packages I opted-out of on deferred-installation (as mentioned above), I wanted every other package to by-default have a deferred-install (provided that I also have by-default `:ensure t`). So in the case of `company` I was expecting that it would be installed on a deferred basis, however that works (my understanding is that `require` is advised so that on the first `require` it prompts the user to install it).

What seems to be happening when I `byte-recompile-directory` is that somehow a `(require 'company)` is occurring and since `company` hasn't been installed yet it fails? The error message I posted a screenshot above seems to show that the `require` advice in the backtrace but I'm not getting any prompt to actually install it.

I think the `require company` is what the `use-package` macro is expanding out to, even though I would expect it shouldn't do that if it indeed correctly assumed that it's by-default `:defer-install t`, so perhaps that's what's failing?

Perhaps of note, I do have `use-package-check-before-init t`. Each time I tested I tested without an `elpa/` dir.

I'll keep looking into it though since you don't seem to be encountering this issue.

blaenk:
Perhaps notably, upon a clean start of the above configuration (with no `elpa/` dir or cache dir or anything), the deferred-packages hash is empty: `#s(hash-table size 65 test eql rehash-size 1.5 rehash-threshold 0.8 data ())`.

blaenk:
I had forgotten that I had already tested without `check-before-init`, and tested again, same results. The `use-package` form keeps expanding out to code that ends up requiring. I even explicitly added `:defer t :ensure t :defer-install t` and it expands to this:

``` elisp
(eval-after-load (quote company) (quote (eval-and-compile (eval-when-compile (with-demoted-errors "Cannot load company: %S" nil (require (quote company) nil (quote nil))))))
```

Later on this ends up expanding into:

``` elisp
(eval-and-compile (eval-when-compile (with-demoted-errors "Cannot load company: %S" nil (require (quote company) nil (quote nil)))))
```

I haven't looked at the use-package macros in a while but it doesn't seem right that it's seemingly unconditionally `require`ing `company` despite a `:defer t :ensure t :defer-install t`, especially since at first it's wrapped in an `eval-after-load 'company`, but then gets expanded as if `company` has indeed been loaded, even though it clearly hasn't because the `require` fails?

blaenk:
Sorry for the spam, but I want to document things as I go.

I think there may be a disconnect between the evaluation environment of my inits and that of `byte-recompile-directory`. For example I get the error mentioned above (regarding a failed `require 'company`) on a clean first-start, and the deferred-packages hash is empty. But if I then go into the `conf/company.el` file and manually macroexpand the `use-package company` form I get:

``` elisp
(progn
  (ignore 'company 't
	  '(:defer-install :ensure :deferred t))
  (puthash 'company
	   '(t :defer-install :ensure :deferred t)
	   use-package--deferred-packages)
  (eval-after-load 'company
    '(eval-and-compile
       (eval-when-compile
	 (with-demoted-errors "Cannot load company: %S" nil
			      (require 'company nil 'nil))))))
```

If I eval-last-sexp that progn, I do get a populated deferred-packages list.

So it seems like things are expanding/evaluating differently when done through byte-recompile-directory.

npostavs:
> So it seems like things are expanding/evaluating differently when done through byte-recompile-directory.

There is some code the purposely produces a different expansion when being byte compiled, e.g., in `use-package-handler/:handler`:
```elisp
    ;; We want to avoid installing packages when the `use-package'
    ;; macro is being macro-expanded by elisp completion (see
    ;; `lisp--local-variables'), but still do install packages when
    ;; byte-compiling to avoid requiring `package' at runtime.
    (if (bound-and-true-p byte-compile-current-file)
        (eval ensure-form)              ; Eval when byte-compiling,
      (push ensure-form body))          ; or else wait until runtime.
```

blaenk:
Thanks, indeed I was just following a lead regarding that after remembering exactly this [from the readme](https://github.com/jwiegley/use-package#prevent-a-package-from-loading-at-compile-time).

> Normally, use-package will load each package at compile time before compiling the configuration, to ensure that any necessary symbols are in scope to satisfy the byte-compiler. At times this can cause problems, since a package may have special loading requirements, and all that you want to use use-package for is to add a configuration to the eval-after-load hook.

So I guess during byte-compilation it tries to `require 'company` to try to pre-compile as much of the code as possible  (?), which goes against/side-steps the defer-install mechanism. Perhaps a separate flag would be required to entirely opt-out of this behavior? Or more out-of-the-box and to the point, opt-out of this behavior if `:defer-install t`. I'm not sure.

blaenk:
This has also alerted me to a misunderstanding I had. Somehow I was expecting that a package could defer-install while also having autoloads automatically created for that package, which would prompt the installation of the package when used, but that's a chicken-and-egg problem since to automatically create the autoloads we would already have to have the package installed, AFAIK. So the alternative to that is to manually create autoloads via the `:commands` etc `use-package` handlers, is that correct?

Given the input:

``` elisp
(use-package company
  :defer t
  :no-require t
  :commands global-company-mode)
```

I macro expanded that like this, thinking maybe it'll trick use-package into thinking that it's being byte-compiled:

``` elisp
(let ((byte-compile-current-file t)) (pp-macroexpand-last-sexp '(4)))
```

I got this output:

``` elisp
(progn
  (ignore 'company 't
	  '(:defer-install :ensure :commands
			   (global-company-mode)
			   :deferred t))
  (puthash 'company
	   '(t :defer-install :ensure :commands
	       (global-company-mode)
	       :deferred t)
	   use-package--deferred-packages)
  (eval-and-compile
    (eval-when-compile
      (with-demoted-errors "Cannot load company: %S" nil nil)))
  (unless
      (fboundp 'global-company-mode)
    (let*
	((load-list-item
	  '(defun . global-company-mode))
	 (already-loaded
	  (member load-list-item current-load-list)))
      (defun global-company-mode
	  (&rest args)
	"[Arg list not available until function definition is loaded.]\n\n(fn ...)"
	(interactive)
	(if
	    (bound-and-true-p use-package--recursive-autoload)
	    (use-package-error
	     (format "Autoloading failed to define function %S" command))
	  (when
	      (use-package-install-deferred-package 'company :autoload)
	    (require 'company)
	    (let
		((use-package--recursive-autoload t))
	      (if
		  (called-interactively-p 'any)
		  (call-interactively 'global-company-mode)
		(apply 'global-company-mode args))))))
      (unless already-loaded
	(setq current-load-list
	      (remove load-list-item current-load-list)))))
  (eval-when-compile
    (declare-function global-company-mode "company")))
```

Then if I eval-last-sexp that expanded form, now I see that the deferred-packages hash is populated and an autoload for global-company-mode exists which upon invocation correctly prompts me to install it.

For some reason this isn't happening when done via `byte-recompile-directory` though. Perhaps my way of tricking use-package's macroexpansion to think it's being byte-compiled is wrong, and what actually gets expanded at byte-compile time is different than what I'm getting here?

raxod502:
I don't have time to debug at the moment, but I can offer a few comments:

* Deferred installation doesn't involve putting any advice on `require`. That was the approach I took in my first commit of #433, but I later decided that it wasn't the right approach (in particular, because autoloads invoked from C code as a result of major mode selection bypass advice). Since the first comment on that thread was just about my first commit, I see the confusion; I've edited in a notice that the approach delineated there is outdated.
* As a consequence of the above, you can't install a deferred package by requiring it. Instead, use `use-package-install-deferred-package`. (That's the reason for the latter function.)
* Unless you specify `use-package-always-ensure` or `use-package-always-defer` (which I highly recommend), `use-package` uses a number of heuristics to sometimes load packages eagerly, and sometimes not. In particular, by default `(use-package company)` is equivalent to `(use-package company :demand t)`.
* What should happen, now that I think of it, is that deferred installation is inhibited (i.e. all packages are installed) when you byte-compile your init-file. That's the default behavior for `use-package`, as far as I can tell from reading the source: packages are installed during compilation. Otherwise macros from the package will not expand correctly.
* I was already curious about the `require` advice in your backtrace, although I didn't mention it because it didn't seem relevant. Because `use-package` certainly does not put any advices on `require`, and I can't think of anything else in Emacs that would do something so aggressive. Are you sure it's not something in your config? (Side note: I Googled the symbol `ad-Advice-require` and found a few seemingly mutually unrelated issue reports, but none of the 111 packages or the Emacs source tree contain references to it. Mysterious indeed!)
* `:demand t` overrides `:defer-install t`. Also, the implied `:demand t` from the heuristic default behavior overrides `:defer-install t`. The latter should probably be changed. But I don't like the heuristic default behavior at all and consider `use-package` unusable without one of `use-package-always-defer` or `use-package-always-demand` set.
* It's unsurprising that the deferred-packages hash is empty. If you specify `:defer-install` and `:demand`, the package is first added to the hash by `:defer-install`, and then installed eagerly when `use-package` processes `:demand`, removing it from the hash. Or that's what should happen. Perhaps there's a more elegant solution, but I don't think it should affect end-user experience.
* I think I might see the issue. `eval-when-compile` forms will be evaluated, as their name suggests, by the byte-compiler no matter where they appear in a form. Even within an `eval-after-load` form. So if deferred-installation packages are not being properly installed during byte-compilation, that could explain the error you're getting.
* That said, I did test with the exact same init-file and auxiliary file that you used. This points to a difference in Emacs version or operating system. You should report these. (In my case, Emacs 25.1 on OS X 10.11.6.) Or it could be some crud that was left over in my `~/.emacs.d`. When I have time, I can do the test again with a clean `~/.emacs.d`.

raxod502:
Oops, Github didn't load additional comments while I was writing my last one. I'll add:

> Somehow I was expecting that a package could defer-install while also having autoloads automatically created for that package, which would prompt the installation of the package when used, but that's a chicken-and-egg problem since to automatically create the autoloads we would already have to have the package installed, AFAIK. So the alternative to that is to manually create autoloads via the :commands etc use-package handlers, is that correct?

Exactly. By definition, there is no way to generate autoloads without doing the very thing that we don't want to do, namely make a network call to download the package. Thus, you have to do everything yourself (if you want to be able to trigger deferred installation with an autoload; after the package is installed, all the autoloads are generated automatically). Add `:commands`, `:mode`, `:interpreter`, etc. This is why I was skeptical about `use-package-always-defer-install`. You have to set things up manually, so it feels more natural for deferred installation to be an opt-in thing.

blaenk:
Thank you for the pointers! I'm going to read through them. The limitations are understandable and to be honest I'm happy to have the option at all to do it this way, so I'm going to give it a shot to try to make it work and report back my findings. I'm willing to put in the effort up-front if it results in a good setup, then I can just keep doing it for each package going forward, which should be simpler.

blaenk:
As for the version, I'm on 25.1.1 x64 from the archlinux repository.

blaenk:
The `require` advice is actually from the `auto-compile` package, which you should have too if you're using the same `init.el` I posted above. That said, in retrospect I should have removed that as it clearly has the potential of muddying the waters when trying to resolve this issue. For the record, I've already tried with it removed.

Information on the advice:

```
ad-Advice-require is a compiled Lisp function.

(ad-Advice-require AD--ADDOIT-FUNCTION FEATURE &optional FILENAME
NOERROR)

Before-advice ‘auto-compile-on-load’:
Before loading the library recompile it if it needs recompilation.
It needs recompilation if it is newer than the byte-compile
destination.  Without this advice the outdated byte-compiled
file would get loaded.

Around-advice ‘build-require-durations’:
Record the time taken to require FEATURE.

[back]
```

Here are the latest narrowed down repros while I comb over your pointers to figure this out:

init.el:

``` elisp
(require 'package)

;; packages
(add-to-list 'package-archives '("melpa" . "http://melpa.org/packages/") t)

(package-initialize)

(unless (package-installed-p 'use-package)
  (package-refresh-contents)
  (package-install 'use-package))

(setq load-prefer-newer t
      backup-by-copying t

      use-package-always-ensure t
      use-package-always-defer-install t
      use-package-check-before-init t
      use-package-enable-imenu-support t)

(require 'use-package)

(defsubst use-package-normalize-value (label arg)
  "Normalize a value."
  (cond ((null arg) nil)
        ((symbolp arg)
         `(symbol-value ',arg))
        ((functionp arg)
         `(funcall #',arg))
        (t arg)))

(require 'diminish)
(require 'bind-key)

(add-to-list 'load-path (expand-file-name "inits/" user-emacs-directory) t)

(byte-recompile-directory (expand-file-name "inits/" user-emacs-directory) 0)
```

inits/conf/company.el

``` elisp
(require 'use-package)

(use-package company)

(provide 'conf/company)
```

backtrace:

backtrace

blaenk:
I forgot to note and mention that benchmark-init also advises require in order to do its profiling. It has also been removed from the above, narrowed repros.

> Unless you specify use-package-always-ensure or use-package-always-defer (which I highly recommend), use-package uses a number of heuristics to sometimes load packages eagerly, and sometimes not. In particular, by default (use-package company) is equivalent to (use-package company :demand t).

I have both always-ensure and always-defer so I guess this isn't the issue, unless they're mutually exclusive, and I don't get the impression that they are.

I'm gonna take a break from this because I'm currently out of ideas. The impression I get as of now is that always-defer is inherently incompatible with byte-compilation, which would be unfortunate, but I expect that I could be wrong given that you don't seem to have any issues.

For the record, each time I test I remove `elpa/` and any other files and folders from `.emacs.d` aside from the init.el and inits/conf/company.el

raxod502:
Oops. I meant `use-package-always-demand`, not `use-package-always-ensure`. Also, I specifically meant `use-package-always-defer` rather than `use-package-always-defer-install`.

I get the same impression as you, but it could be possible to make it work. Also, I'll (at some point) retry my test. I may have forgotten to wipe `elpa` (the thought didn't occur to me since I don't use `package.el`).

blaenk:
Ahh okay that explains things. I'll have to look into that.

> I think I might see the issue. eval-when-compile forms will be evaluated, as their name suggests, by the byte-compiler no matter where they appear in a form. Even within an eval-after-load form. So if deferred-installation packages are not being properly installed during byte-compilation, that could explain the error you're getting.

This seems to be what's going on based on the backtrace. It'll show an expansion of a form and then expand a specific `eval-when-compile` within it. It does appear like byte-compilation isn't causing the packages to be installed.

If byte-compiling is indeed inherently incompatible with defer-install because byte-compilation would require the package to be installed, then I guess it's a matter of weighing each optimization. On the one hand defer-install would allow me to keep my elpa dir small until I actually use certain packages, but I'll miss out on the overall byte compilation of my init files. In the extreme eventual case where every defer-install package is eventually installed, I'll no longer reap the benefits of that "optimization", though at that point I could safely flip the switch on byte compilation because the packages will exist.

That leads me to wonder if perhaps some kind of workaround would be reasonable where a :defer-install package would opt itself out of byte-compilation in some way? Is that even possible? That way the interior eval-when-compile forms won't be evaluated. Then my manually-defined autoloads via e.g. `:commands` can take charge of eventually installing the package, then once installed it would no longer opt-itself out of byte-compilation?

Perhaps that is more trouble than is worth, if it's even possible. There's no rush from my end, for now I'll revert to not using `:defer-install` until I think through what I want to do and give you some time to see how things fit.

raxod502:
Well, I don't byte-compile my init-files, so I'm not overly familiar with all the potential issues. Emacs certainly *wants* you to load your packages during byte-compilation for things like e.g. macroexpansion, but you don't *have* to. It would be a horrifying hack, but I can imagine saving the user's `:config` section as an unevaluated form and then calling `byte-compile` on it manually once the package is installed. Then again deferred installation as a whole is kind of already a horrifying hack :)

This is an interesting implementation question that will only be properly answerable once my email volume drops below 100 per day (just counted, it's a new record!) ;)

raxod502:
I thought about this some more, and concluded that it should be possible, provided the following steps are taken:

* The interface for `use-package-ensure-function` and `use-package-pre-ensure-function` is changed so that the functions are called during macroexpansion instead of at runtime, and they return an unevaluated form it install the package, rather than actually installing the package.
* Create the `use-package--deferred-packages` variable in `use-package-handler/:ensure` if it does not exist, since the `puthash` is evaluated at runtime when `use-package` may not be present.
* Autoload `use-package-install-deferred-package`.
* Don't use macros from packages that have deferred installation, because these won't work. (They could be made to work, but it would be harder.)
* Test to make sure this actually works, because I'm not necessarily convinced it does.

I don't have time at the moment to implement this but would be interested in others' opinions.

PR #447, opened 2017-04-03
Don't mutilate keyword arguments in :bind

raxod502:
The parsing logic in `use-package-normalize-pairs` is not designed to
deal with keyword arguments. However, `use-package-normalize-pairs` is
used to process the arguments to :bind, which can include keyword
arguments. These keyword arguments are supposed to be passed untouched
to the underlying `bind-keys` function, but there is a clause in
`use-package-normalize-pairs` that replaces lists with their first
element. Thus an invocation like:

```
(use-package company
  :bind (:map company-active-map
         :filter (company-explicit-action-p)
         ("RET" . company-complete-selection)))
```

Generates code like this:

```
(bind-keys
  :map company-active-map
  :filter company-explicit-action-p
  ("RET" . company-complete-selection))
```

Which generates an error since `company-explicit-action-p` is now
being referenced as a variable rather than a function.

The proper solution is to refactor the logic that goes into parsing
uses of :bind, but this commit adds a temporary patch to eliminate the
above problem, while trying to be as reverse-compatible as possible.
In particular it just inhibits the list-to-first-element
transformation when the previous element processed was a keyword.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -964,12 +964,15 @@ If RECURSED is non-nil, recurse into sublists."
    ((use-package-is-pair arg key-pred val-pred)
     (list arg))
    ((and (not recursed) (listp arg) (listp (cdr arg)))
-    (mapcar #'(lambda (x)
-                (let ((ret (use-package-normalize-pairs
-                            key-pred val-pred name label x t)))
-                  (if (listp ret)
-                      (car ret)
-                    ret))) arg))
+    (let ((last-item nil))
+      (mapcar #'(lambda (x)
+                  (prog1
+                      (let ((ret (use-package-normalize-pairs
+                                  key-pred val-pred name label x t)))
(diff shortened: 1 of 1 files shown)


npostavs:
> The proper solution is to refactor the logic that goes into parsing
> uses of :bind, but this commit adds a temporary patch to eliminate the
> above problem,

That seems to me like an acceptable hack, but it should come with a comment in the code to explain that.

jwiegley:
Agreed; can you please add a comment, along with your description of the proper fix to be made in the future? Then I'm happy to merge it.

raxod502:
Added. As I say in the comment, the long-term solution will require a reconciling of the inconsistent handling of keyword arguments in `use-package` vs. `bind-key`. (Actually, I don't think it's necessarily a bad thing that the behavior is different—but we need to decide what is desired, and then document it.)

PR #451, opened 2017-04-10
More complex conditional example

keithmantell:
As a bit of an elisp newbie I liked this more complex example!

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,7 @@ In another example, we can load things conditional on the operating system:
 
 ```
 (use-package exec-path-from-shell
-  :if (string-equal system-type "darwin")
+  :if (eq system-type 'darwin)
   :ensure t
   :config
   (exec-path-from-shell-initialize))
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -245,6 +247,15 @@ graphical Emacs, not for other Emacsen I may start at the command line:
   (add-hook 'after-init-hook 'server-start t)
   (add-hook 'after-init-hook 'edit-server-start t))
 ```
+In another example, we can load things conditional on the operating system:
+
+```
+(use-package exec-path-from-shell
+  :if (string-equal system-type "darwin")
```
npostavs: Use `(eq system-type 'darwin)` though.

keithmantell: Great! I did say i was a newbie!

New PR required?

npostavs: This is a very minor change, so IMO it should be okay to amend the commit and then `push --force` to update this PR.

jwiegley:
Thanks!

nickmccurdy:
Please see #452, I updated the predicate to the recommended one that detects if Emacs is running in a terminal (in which case you don't need to load `exec-path-from-shell`).

PR #463, opened 2017-05-23
Protect against errors during package install

leathekd:
If the network is missing and there is a new use-package with :ensure,
startup would fail part of the way through due package.el being unable
to reach the package repo.  This will catch that error and report it
while allowing startup to continue.

Discovered and fixed on an airplane.  :)


leathekd:
Thanks!

PR #467, opened 2017-05-30
Use Cask and ert-runner

nickmccurdy:
Thanks to Cask, dependencies will now be installed automatically in CI (you can also run `cask install` to do this locally). I also added ert-runner for more standard test execution, while making it easier to add more test files.


test/use-package-test.el
```diff
@@ -1,4 +1,4 @@
-;;; use-package-tests.el --- Tests for use-package.el
+;;; use-package-test.el --- Tests for use-package.el
```
npostavs: Hmm, the Emacs convention is `<pkgname>-tests.el` not `-test.el`, does cask require this name change?

nickmccurdy: I named it with the `-test.el` suffix because that was the exact filename generated by ert-runner's init feature. It might work without this rename but I haven't tried it yet (and the file would still have to be in the `test` directory for ert-runner). Also I thought `-test.el` was the convention but I could be wrong.

npostavs: The test files that come with Emacs all use `-tests.el`(e.g., http://git.savannah.gnu.org/cgit/emacs.git/tree/test/lisp). It's too bad that the tools outside of Emacs decided on a different convention.

nickmccurdy: Thanks for the example, I thought it would be a mix of both suffixes. Since this might eventually be integrated into Emacs, I'll see if ert-runner takes the `-tests.el` suffix.

nickmccurdy: Unfortunately ert-runner sees no tests when I use the `-tests.el` suffix, so I'm reverting this rename. The options are to use the `-test.el` suffix or not use ert-runner.

https://travis-ci.org/jwiegley/use-package/builds/239172774#L569

Issue #288, opened 2015-11-22
diminish imply commands


fommil:
it would be really great if `:diminish` implied `:commands` as I find myself repeating the minor mode in both these sections when I want to lazily load.

npostavs:
But then it would be impossible to `:diminish` without loading lazily, right?

fommil:
ah

jwiegley:
You can always use `:demand t` to avoid lazy loading.

Looking at my own `init.el`, I too repeat the major mode twice nearly everywhere that I use `:diminish`.  The only problem with making this change is that it might surprise many of its current users...

npostavs:
> You can always use :demand t to avoid lazy loading.

Ah, good point.

> Looking at my own init.el, I too repeat the major mode twice nearly everywhere that I use :diminish.

Hah, I've looked at mine, and I found a few places where I used `:diminish` without `:defer`, but in each of those cases I see that I *should* be using `:defer` anyway!

> The only problem with making this change is that it might surprise many of its current users...

Yes, in a few of the cases mentioned above, I'd also need to move some code from `:config` to `:init` (though again, I should probably do this anyway).

But still, on the whole, +1 from me on this change.

jwiegley:
Given the different ways that `:commands` and `:diminish` are handled in the code, I'm going to avoid making this any more magical.  It will simply require duplication if lazy loading is desired.

jwiegley:
Something we might want someday is the ability to have "plugin optimizers", or code passes that are free to rewrite `use-package` declarations according to rules specified by the user.  Then request like this would be trivial, as they would happen before we start expanding the macro.

PR #472, opened 2017-06-14
Allow `:diminish` with no arguments.

joewreschnig:
When given no arguments, have :diminish assume it should diminish a
mode named after the current package (the package’s name, with “-mode”
appended, if it’s not already) to an empty string.

When given only a string to diminish an implicit package name to, do
not append “-mode” to the package name if it already ends with
it.  (This is a backwards-incompatible change if someone was
diminishing a package named “foo-mode” implementing `foo-mode-mode`.)

Add test cases for `use-package-normalize-diminish`.

This addresses some of the redundancy mentioned in issue #288.

### DIFF
--- a/use-package-tests.el
+++ b/use-package-tests.el
@@ -72,6 +72,18 @@
 
   )
 
+(ert-deftest use-package-normalize-diminish ()
+  (should (equal (use-package-normalize-diminish 'foopkg :diminish nil)
+                 '(foopkg-mode)))
+  (should (equal (use-package-normalize-diminish 'foopkg :diminish 'bar)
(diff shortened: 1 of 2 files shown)


nickmccurdy:
👍 I assumed that use-package was supposed to work like this and eventually gave up after using `:diminish` without arguments seemed to not do anything. Thank you.

joewreschnig:
Yes, I had a similar experience with `:diminish t` (which I probably assumed because of `:ensure t`). But I think nothing is even better than `t`.

jwiegley:
We should allow the same for `:disabled`.

PR #479, opened 2017-06-30
Fix bug in use-package-install-deferred-package

raxod502:
Previously, deferred installation didn't work because I didn't convert the result of a `completing-read` back from a string to a symbol, which meant the hash-table lookup failed.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -669,11 +669,12 @@ If the package is installed, its entry is removed from
               use-package--deferred-packages)
      (if packages
          (list
-          (completing-read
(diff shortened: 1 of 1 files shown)


raxod502:
To clarify, this will not be a breaking change, unless somebody was relying on the previous behavior of `use-package-install-deferred-package` to silently do nothing no matter what package was selected.

raxod502:
@jwiegley any news here?

raxod502:
Thanks!

Issue #477, opened 2017-06-25
:delight is inflexible and verbose


joewreschnig:
I'm switching my configuration from `:diminish` to `:delight` because I want a few `:eval` forms. Right now, `:delight` is both limited - I can't delight multiple modes - and verbose - I have to specify mode names ([which I don't like doing](https://github.com/jwiegley/use-package/pull/472)). It's also a little annoying to have to quote the second argument, since `use-package` keyword values usually don't need to be quoted.

Some of the forms I'd like:

* **Just hide it:**
`(use-package foo-mode :delight)` to `(delight 'foo-mode nil 'foo-mode)`
* **Assume `-mode`**:
`(use-package foo :delight)` to `(delight 'foo-mode nil 'foo)`
* **Only a string**:
`(use-package foo-mode :delight "F")` to `(delight 'foo-mode "F" 'foo-mode)`
`(use-package foo :delight "F")` to `(delight 'foo-mode "F" 'foo)`
* ~~**Unquoted :eval**:~~ This is not possible if multiple delights are required, as it's ambiguous for any symbol other than `:eval`.
* **Multiple calls:**
`(use-package foo :delight (foo-mode "F") (foo-global-mode "fg")` to
`(delight 'foo-mode "F" 'foo)`
`(delight 'foo-global-mode "fg" 'foo)`

joewreschnig:
Unfortunately mode lines support a form `(sym a)` and `(sym a b)` to display `a` if `sym` is non-nil and `b` otherwise. This introduces some ambiguity when supporting multiple forms.

`(use-package foo :delight (bar-mode "x") ...` could mean `(delight 'foo-mode '(bar-mode "x") 'foo)` i.e. light `foo-mode` as "x" when `bar-mode` is set, or it could mean `(delight 'bar-mode "x" 'foo)` i.e. package `foo` contains a mode named `bar-mode` which should be delighted to "x".

PR #480, opened 2017-07-02
Allow multiple :delight arguments, or omitting the mode. (#477)

joewreschnig:
This allows using forms such as

    (use-package foo :delight)
        ;; => (delight 'foo-mode)
    (use-package foo :delight " f")
        ;; => (delight 'foo-mode " f")
    (use-package foo :delight (a-mode) (b-mode " b")
        ;; => (delight 'a-mode) (delight 'b-mode " b")

This brings support for `:delight` in line with `:diminish`.


joewreschnig:
The first patch retained the original code's forced quoting of the third argument, which is not correct when the third argument is `:major`. This is handled correctly now. As a result all :delight forms for one package are compressed into one `(delight ...)` call, which is probably nicer to read in most cases.

Issue #481, opened 2017-07-04
Document :delight keyword in the README


Silex:
:-)

ping @darkfeline

joewreschnig:
I can do this, if it wouldn't be stepping on anyone's toes...

Silex:
@joewreschnig: I don't think anyone will object

joewreschnig:
I don't think the last example is very good; people are likely to get hung up on the `emacs` package, which isn't referred to much. But it's the only time in my init.el I delight two modes for one package; it's rare to even have two modes in a package, let alone want to delight both.

Silex:
:+1: 

Closing, next comments will be on the PR

PR #484, opened 2017-07-09
Document :delight in the doc string and README.

joewreschnig:
This addresses issue #481.

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,15 +350,17 @@ looking up the same information again on each startup:
   :commands R)
 ```
 
-## Diminishing minor modes
+## Diminishing and delighting minor modes
 
-`use-package` also provides built-in support for the diminish utility -- if
-you have that installed.  Its purpose is to remove strings from your mode-line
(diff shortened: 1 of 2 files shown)


joewreschnig:
I don't think the last example is very good; people are likely to get hung up on the emacs package, which is unusual if you haven't seen it before. But it's the only time in my init.el I delight two modes for one package; it's rare to even have two modes in a package, let alone want to delight both.

README.md
```diff
@@ -368,6 +370,38 @@ end:
...
+;; Remove the mode name for projectile-mode, but show the project name.
+(use-package projectile
+  :delight '(:eval (concat " " (projectile-project-name))))
+
+;; Completely hide visual-line-mode and change auto-fill-mode to " AF".
+(use-package emacs
+  :delight
+  (auto-fill-function " AF")
+  (visual-line-mode))
+
```
Silex: The extra blank line here should be removed IMHO

Silex:
Thanks @joewreschnig!

PR #508, opened 2017-10-17
Add customization keywords

canatella:
Hello,

This PR adds two new keyword: `:customize` and `:custom-faces`. These are shortcuts to cleanup calls to `customize-set-variable` and `custom-set-faces'`in the `:config` part. I do not use the Emacs internal save to custom.el setup because it's not easy to track in version control. I prefer to keep my package configuration in one place and these two keywords make this easier.

``` elisp
(use-package comint
  :customize
  (comint-buffer-maximum-size 20000 "Increase comint buffer size.")
  (comint-prompt-read-only t "Make the prompt read only."))

(use-package eruby-mode
  :custom-faces
  (eruby-standard-face ((t (:slant italic)))))
```

The comments are not mandatory for the `:customize` keyword.

I don't know if you would be willing to add these kind of keywords, so this PR is more about asking if theses features would be of interest for you to merge as I don't think this PR can be merged as is. I'm not sure about the normalizing function. They do not look like the other one because I did not really understood them, so I simply wrote ones that did the trick.

### DIFF
--- a/use-package.el
+++ b/use-package.el
@@ -159,6 +159,7 @@ the user specified."
     :functions
     :defer
     :custom
+    :custom-face
     :init
     :after
     :demand
@@ -1441,7 +1442,34 @@ deferred until the prefix key sequence is pressed."
              args)
      body)))
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;;; :custom-face
+;;
 
+(defun use-package-normalize/:custom-face (name-symbol keyword arg)
+  "Normalize use-package custom-face keyword."
+  (let ((error-msg (format "%s wants a (<symbol> <face-spec>) or list of these" name-symbol)))
+    (unless (listp arg)
+      (use-package-error error-msg))
+    (dolist (def arg arg)
+      (unless (listp def)
+        (use-package-error error-msg))
+      (let ((face (nth 0 def))
+            (spec (nth 1 def)))
+        (when (or (not face)
+                  (not spec)
+                  (> (length arg) 2))
+          (use-package-error error-msg))))))
+
+(defun use-package-handler/:custom-face (name keyword args rest state)
+  "Generate use-package custom-face keyword code."
+  (let ((body (use-package-process-keywords name rest state)))
+    (use-package-concat
+     (mapcar (lambda (def)
+               `(custom-set-faces (quote ,def)))
+             args)
+     body)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;
@@ -1590,6 +1618,7 @@ this file.  Usage:
 :diminish        Support for diminish.el (if installed).
 :delight         Support for delight.el (if installed).
 :custom          Call `customize-set-variable' with each variable definition.
+:custom-face     Call `customize-set-faces' with each face definition.
 :ensure          Loads the package using package.el if necessary.
 :pin             Pin the package to an archive."
   (declare (indent 1))
--- a/use-package.el
+++ b/use-package.el
@@ -158,6 +158,7 @@ the user specified."
     :defines
     :functions
     :defer
+    :custom
     :init
     :after
     :demand
@@ -1406,6 +1407,44 @@ deferred until the prefix key sequence is pressed."
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;
+;;; :custom
+;;
+
+(defun use-package-normalize/:custom (name-symbol keyword arg)
+  "Normalize use-package custom keyword."
+  (let ((error-msg (format "%s wants a (<symbol> <form> <optional string comment>) or list of these" name-symbol)))
+    (unless (listp arg)
+      (use-package-error error-msg))
+    (dolist (def arg arg)
+      (unless (listp def)
+        (use-package-error error-msg))
+      (let ((variable (nth 0 def))
+            (value (nth 1 def))
+            (comment (nth 2 def)))
+        (when (or (not variable)
+                  (not value)
+                  (> (length def) 3)
+                  (and comment (not (stringp comment))))
+          (use-package-error error-msg))))))
+
+(defun use-package-handler/:custom (name keyword args rest state)
+  "Generate use-package custom keyword code."
+  (let ((body (use-package-process-keywords name rest state)))
+    (use-package-concat
+     (mapcar (lambda (def)
+               (let ((variable (nth 0 def))
+                     (value (nth 1 def))
+                     (comment (nth 2 def)))
+                 (unless comment
+                   (setq comment (format "Customized with use-package %s" name)))
+                 `(customize-set-variable (quote ,variable) ,value ,comment)))
+             args)
+     body)))
+
+
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
 ;;; :diminish
 ;;
 
@@ -1550,6 +1589,7 @@ this file.  Usage:
 :load-path       Add to the `load-path' before attempting to load the package.
 :diminish        Support for diminish.el (if installed).
 :delight         Support for delight.el (if installed).
+:custom          Call `customize-set-variable' with each variable definition.
 :ensure          Loads the package using package.el if necessary.
 :pin             Pin the package to an archive."
   (declare (indent 1))
--- a/README.md
+++ b/README.md
@@ -205,6 +205,21 @@ This does exactly the same thing as the following:
   :bind ("C-." . ace-jump-mode))
 ```
 
+## Package customization
+
+### Customizing variables.
+
+The `:custom` keyword allows customization of package custom variables.
+
+``` elisp
+(use-package comint
+  :custom
+  (comint-buffer-maximum-size 20000 "Increase comint buffer size.")
+  (comint-prompt-read-only t "Make the prompt read only."))
+```
+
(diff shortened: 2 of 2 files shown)


jwiegley:
OK, this is a pretty brilliant idea, thank you! Have you signed copyright papers to the FSF?  That's now a requirement, since we're soon to merge into Emacs core.

canatella:
Nope, where should I start for the assignment ? Should I ask on emacs-devel as suggested in the Emacs manual ?

jwiegley:
Send an e-mail to assign@gnu.org and they'll get you sorted out. Thanks!

use-package.el
```diff
@@ -1406,6 +1408,67 @@ deferred until the prefix key sequence is pressed."
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;
+;;; :customize
+;;
+
+(defun use-package-normalize/:customize (name-symbol keyword arg)
+  "Normalize use-package customize keyword."
+  (let ((error-msg (format  "%s wants a (<symbol> <form> <optional string comment>) or list of these" name-symbol)))
```
basil-conto: Nitpick: extraneous space after `format`.

use-package.el
```diff
@@ -1406,6 +1408,67 @@ deferred until the prefix key sequence is pressed."
...
+    (unless (listp arg)
+      (use-package-error error-msg))
+    (dolist (def arg arg)
+      (unless (listp def)
+        (use-package-error error-msg))
+      (seq-let [variable value comment] def
```
basil-conto: Emacs 24 and older do not include `seq.el`.

use-package.el
```diff
@@ -1406,6 +1408,67 @@ deferred until the prefix key sequence is pressed."
...
+                  (not value)
+                  (> (length def) 3)
+                  (and comment (not (stringp comment))))
+          (use-package-error error-msg))))))
+
+(defun use-package-handler/:customize (name keyword args rest state)
+  "Generate use-package customize keyword code."
+  (let ((body (use-package-process-keywords name rest state)))
+    (use-package-concat
+     (seq-map (lambda (def)
```
basil-conto: Assuming `seq.el` is available, is there a reason to prefer `seq-map` over `mapcar` here?

use-package.el
```diff
@@ -1406,6 +1408,67 @@ deferred until the prefix key sequence is pressed."
...
+     body)))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;
+;;; :custom-faces
+;;
+
+(defun use-package-normalize/:custom-faces (name-symbol keyword arg)
+  "Normalize use-package custom-faces keyword."
+  (let ((error-msg (format  "%s wants a (<symbol> <face-spec>) or list of these" name-symbol)))
```
basil-conto: Ditto re: extraneous space.

use-package.el
canatella: Are there some other form of destructuring binding I could use ?

use-package.el
canatella: When I finally have a nice sequence elisp library, I cannot use it anymore ;) I'm just using the seq-* function in my elisp because they seem more modern then the old disparate functions, but I understand the point here and will switch to mapcar.

use-package.el
dieggsy: you could  try `cl-destructuring-bind`?

canatella:
Ok, so I

- replaced seq-map by mapcar,
- replaced seq-let by cl-destructuring bind,
- and removed the extra spaces.

I also contacted assign@gnu.org for the paper work.

use-package.el
```diff
@@ -1406,6 +1408,67 @@ deferred until the prefix key sequence is pressed."
...
+
+(defun use-package-normalize/:customize (name-symbol keyword arg)
+  "Normalize use-package customize keyword."
+  (let ((error-msg (format "%s wants a (<symbol> <form> <optional string comment>) or list of these" name-symbol)))
+    (unless (listp arg)
+      (use-package-error error-msg))
+    (dolist (def arg arg)
+      (unless (listp def)
+        (use-package-error error-msg))
+      (cl-destructuring-bind (&optional variable value comment) def
```
basil-conto: `cl-destructuring-bind` is a macro defined in `cl-macs.el`, so you will need to `(eval-when-compile (require 'cl-macs))`. I do not, however, think this is the best destructuring option here, as it will throw a non-`use-package`-specific `wrong-number-of-arguments` error if `def` has more elements than it is being destructured into (this also makes the `(> (length def) 3)` check further down redundant).

I suggest using `pcase-let` instead:

```el
(pcase-let ((`(,variable ,value ,comment) def))
  ;; ...
  )
```

This has the added benefit of introducing no further dependencies (apart from the already prevailing Emacs24+ requirement).

npostavs: > `(eval-when-compile (require 'cl-macs))`

FYI, it should rather be `(eval-when-compile (require 'cl-lib))` (e.g., see https://lists.gnu.org/archive/html/emacs-devel/2014-01/msg01003.html).

> I suggest using `pcase-let` instead:

Hmm, I'm not sure that one deals well with a different number of arguments either:

```
(pcase-let BINDINGS &rest BODY)

Like ‘let’ [...]
The macro is expanded and optimized under the assumption that those
patterns *will* match, so a mismatch may go undetected or may cause
any kind of error.
```

basil-conto: I always feared that might be the case re: `cl-macs` so thanks for confirming it. :)

My reading of the `pcase` and `pcase-let` documentation is that a backquote-style `QPAT` list pattern will always match, regardless of the length of the list being matched against. In other words, I think the `pcase-let` docstring is warning against uses like `(pcase-let (((and (pred listp) l) "wut")))`, which don't match but which `pcase` would happily skip.

Maybe I'm wrong about this, and maybe destructuring isn't worth it for such simple validation checks.

npostavs: > a backquote-style QPAT list pattern will always match, regardless of the length of the list being matched against.

Well if that were true, how would you distinguish between different list lengths? Perhaps noting that `` `(,one ,two ,three)`` is a shorthand for `` `(,one ,two ,three . nil)`` might make it clearer:

```elisp
(mapcar (lambda (x)
          (pcase x
            (`(,one ,two ,three . nil) (list :exactly-three-args one two three))
            (`(,one ,two ,three . ,rest) (list :more-than-three-args one two three rest))
            (_ (list :less-than-three-args x))))
        '((1) (1 2) (1 2 3) (1 2 3 4)))
;=>
((:less-than-three-args
  (1))
 (:less-than-three-args
  (1 2))
 (:exactly-three-args 1 2 3)
 (:more-than-three-args 1 2 3
                        (4)))
```

canatella: Going with

``` elisp
(let ((var (car def))
      (val (cadr def))
      (com (caadr def)) ...)
```
then ?

npostavs: You mean `(caddr def)`, but otherwise seems fine.

canatella: indeed, that proves what's nice about destructuring bind ;)

canatella:
I replaced the destructuring bind with a good old let.

use-package.el
basil-conto: @npostavs I see, thanks for clarifying.

@canatella I don't think Emacs 24 has `caddr` yet. Why not just use `nth` for all of them? That way you afford a tiny bit more destructional clarity.

canatella:
Indeed, `nth` keeps a bit of the destructuring clarity. Replaced `ca*dr` with `nth`.

canatella:
Hello,

No answer from assign@gnu.org for 2 days, does it always take a bit of time, or should I try contacting them again ?

canatella:
Another thing, wouldn't it be better to customize at `:init` rather then `:config` ? I have a problem here with `exec-path-from-shell` which needs the `exec-path-from-shell-variables` to be setup before calling `exec-path-from-shell-initialize`.

This is not working with the current code:

``` elisp
(use-package exec-path-from-shell
  :ensure t
  :customize
  (exec-path-from-shell-variables
   '("PATH" "MANPATH" "ANDROID_HOME" "ANDROID_NDK_HOME"))
  :config
  (when (memq window-system '(mac ns x))
    (exec-path-from-shell-initialize)))
```

While this is

``` elisp
(use-package exec-path-from-shell
  :ensure t
  :customize
  :config
  (when (memq window-system '(mac ns x))
    (let ((exec-path-from-shell-variables
           '("PATH" "MANPATH" "ANDROID_HOME" "ANDROID_NDK_HOME")))
      (exec-path-from-shell-initialize))))
```
works as expected.

npostavs:
> No answer from assign@gnu.org for 2 days, does it always take a bit of time, or should I try contacting them again ?

I think it was something like 30 days when I did it.

canatella:
Well, let's wait then...

alphapapa:
I did the CA recently and got responses within 24 hours IIRC.  But AFAIK there's only one person who does that job, so I guess he's busier at some times than others.

canatella:
Hello, the copyright assignment procedure is done, so we can go on but I have discovered a problem in my config though.

As is, this patch sets the custom variables at `:config` time. It's a problem for the `exec-path-from-shell` package which expects the variable `exec-path-from-shell-variables` to be set before it is required. Is it a problem with the package, or should the variables be set at `:init` time instead?

jwiegley:
I believe they should be set at `:init` time, as regular customizations are.

canatella:
I've modified the keyword order so that the customizations are done just before `:init`. The `exec-path-from-shell` package is now working correctly.

dieggsy:
@canatella should `:customize` be `:custom-variables` for consistency? (I would even consider `:custom-set-variables` and `:custom-set-faces` for clarity, though those are pretty wordy

canatella:
I asked myself that question too. While I agree with you on the need for keyword consistency, I think that the `:custom-face' one would be much less used and keeping keyword length reasonable is nice.

What do other reviewers here think ?

jwiegley:
I prefer `:custom` and `:custom-face`.

canatella:
Good compromise, I made the change.

dieggsy:
@jwiegley @canatella any news on this?

jwiegley:
Thank you for a great addition!

PR #521, opened 2017-11-29
Require cl-lib and do not use seq functions

### LINKED ISSUE #520: use-package-normalize/:hook requires seq-every-p
... which results in void-function seq-every-p because there's no `(require 'seq)`

basil-conto:
`cl-lib` is also required for a separate call to `cl-mapcan`.

Fixes #520


jwiegley:
Thanks!

PR #552, opened 2017-12-05
Correct value of use-package-defaults for :ensure

raxod502:
See https://github.com/raxod502/straight.el/issues/193.

### DIFF
--- a/use-package-ensure.el
+++ b/use-package-ensure.el
@@ -195,10 +195,11 @@ manually updated package."
     body))
 
 (add-to-list 'use-package-defaults
-             '(:ensure (list use-package-always-ensure)
-                       (lambda (args)
-                         (and use-package-always-ensure
-                              (not (plist-member args :load-path))))) t)
+             '(:ensure use-package-always-ensure
+               (lambda (args)
+                 (and use-package-always-ensure
+                      (not (plist-member args :load-path)))))
+             t)
 
 (add-to-list 'use-package-defaults
              '(:pin use-package-always-pin use-package-always-pin) t)


jwiegley:
@raxod502 Can you please explain this change by way of a test? If not, I'll try to make one until I understand the change before merging.

`use-package-defaults' should probably be changed to use *pre-normalization* values, although right now it's using *post-normalization* values. In the case of `:ensure`, the post-normalization form is a list, which is why I was using a list.

raxod502:
Yes, and I'm sorry for not providing any reasoning to begin with. I've now taken another look at the code.

It seems to me that the `:pin` support for `:ensure` is currently implemented in `use-package-normalize/:ensure`. I think it would make more sense for it to be implemented in `use-package-ensure-elpa`. This is because `:ensure` will be used for all package managers, but `:pin` is only relevant to `package.el`.

In fact, I think it would make the most sense to pass the entire argument of `:ensure` directly to `use-package-ensure-function`, unmodified. Some package managers might have a richer package description syntax that would include data structures other than lists and symbols, for example, and we wouldn't want to exclude those.

Does that sound reasonable to you? If so, I can produce patches that will implement this and make `straight.el` compatible with it.

jwiegley:
Yes, your renewed proposal sounds like the right solution. I look forward to your revised PR.

jwiegley:
Closing this PR, in expectation of a new one once the matter has been resolved in `straight.el`.

raxod502:
Well, it's not an issue to be resolved in `straight.el`. But I agree this PR should be closed since the proper patch to `use-package` looks nothing like this one. I'll make a new one.

PR #576, opened 2017-12-07
Make use-package.org ox-hugo export ready

kaushalmodi:
This commit adds the code to make the doc-site of use-package happen
by using emacs + Org + ox-hugo + hugo on CI/CD.

Once Travis is set up, commits made to the use-package.org will
auto-update the doc site in few minutes.

Content creation and changes in doc/content/ is ignored. That content
is auto-generated on CI/CD and must not be committed.

### Workaround to allow outside links to work for texinfo & hugo exports

A new Org macro link-jump is added so that both texinfo and hugo
exports work harmoniously.

For texinfo exports, the export is done using the scope of the whole
file, so it can resolve all internal link references.

Whereas for hugo exports, they are done only from the scope of a
subtree (or a page of the doc site), so at the moment it doesn't
auto-resolve Org internal links outside that scope.

    #+MACRO: link-jump @@texinfo:@ref{$1}@@@@hugo:[[$2][$1]]@@

Example use:

    {{{link-jump(FAQ,/faq)}}}


doc/Makefile
```diff
@@ -0,0 +1,169 @@
...
+EMACS ?= emacs
+EMACS_exists := $(shell command -v $(EMACS) 2> /dev/null)
+ifeq ("$(EMACS_exists)","")
+	EMACS := /tmp/emacs/bin/emacs
+endif
+
+# EMACS_BIN_SOURCE and EMACS_VERSION are used later in the vcheck rule
+# only if EMACS_exists has evaluated to "".
+EMACS_BIN_SOURCE ?= https://github.com/npostavs/emacs-travis/releases/download/bins
+EMACS_VERSION ?= 25.2
```
npostavs: You should use 25.3, there's never any reason to use 25.2 when 25.3 is available.

kaushalmodi: Relic of the past, will fix it. Thanks!

PR #583, opened 2017-12-08
Move EMACS_VERSION=26 to the top so that the docs build first

kaushalmodi:
The order of the emacs versions in the matrix is changed so that the CI job with emacs 26 runs first.

The doc site is generated using emacs 26, and so the doc site will be updated first. 
Other emacs versions are ordered so that the newer versions are tested first and then the older.

The master version of emacs is still kept at the last.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,11 @@ env:
   global:
     - CURL="curl -fsSkL --retry 9 --retry-delay 9"
   matrix:
-  - EMACS_VERSION=24.3
-  - EMACS_VERSION=24.4
-  - EMACS_VERSION=24.5
-  - EMACS_VERSION=25.3
   - EMACS_VERSION=26
+  - EMACS_VERSION=25.3
+  - EMACS_VERSION=24.5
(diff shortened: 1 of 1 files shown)


nickmccurdy:
Does Travis run the deployment as soon as the current build passes or does it wait for the suite to pass?

kaushalmodi:
The deployment step is anyways at the end: https://travis-ci.org/jwiegley/use-package/jobs/313701462

So I believe the deployment won't happen if the test fails..

But as soon as the current build passes, the deployment finishes.. as it has nothing to do with the tests for other emacscen versions in the matrix.

jwiegley:
From my watching of Travis, it was building all the various Emacsen versions simultaneously.

kaushalmodi:
I saw that only the first 5 started at almost the same time. The 6th started after 1 of 5 finished. But to avoid any chance, moved Emacs 26 to the top.

kaushalmodi:
The major speedup was of course by the prior commit, that switched to using Emacs 26 for the doc site.

PR #584, opened 2017-12-09
Pass more information to :ensure function

raxod502:
This pull request supersedes https://github.com/jwiegley/use-package/pull/552.

@jwiegley Can you confirm that this seems reasonable? If so, I will go ahead and update the test suite (which is currently failing).

### DIFF
--- a/use-package-ensure.el
+++ b/use-package-ensure.el
@@ -56,7 +56,7 @@ See also `use-package-defaults', which uses this value."
   :type 'symbol
   :group 'use-package-ensure)
 
-(defcustom use-package-ensure-function 'use-package-ensure-elpa
+(defcustom use-package-ensure-function #'use-package-ensure-elpa
   "Function that ensures a package is installed.
 This function is called with three arguments: the name of the
 package declared in the `use-package' form; the arguments passed
@@ -133,10 +133,10 @@ manually updated package."
 (defvar package-archive-contents)
 
 ;;;###autoload
-(defun use-package-normalize/:ensure (name keyword args)
+(defun use-package-ensure-elpa-normalize (args)
   (if (null args)
       (list t)
-    (use-package-only-one (symbol-name keyword) args
+    (use-package-only-one "ensure" args
       #'(lambda (label arg)
           (cond
            ((symbolp arg)
@@ -153,6 +153,7 @@ manually updated package."
                      "(an unquoted symbol name), or (<symbol> :pin <string>)"))))))))
 
 (defun use-package-ensure-elpa (name args state &optional no-refresh)
+  (setq args (use-package-ensure-elpa-normalize args))
   (dolist (ensure args)
     (let ((package
            (or (and (eq ensure t) (use-package-as-symbol name))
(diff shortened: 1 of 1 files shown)


jwiegley:
I don't quite understand what this change aims to do. At the very least, you need `use-package-normalize/:ensure`, otherwise it's an error for the keyword's normalizer to be missing. Should this function just be calling the new one you've defined?

raxod502:
@jwiegley What this change aims to do is have the normalization take place _inside `use-package-ensure-elpa`_, rather than in the normalizer. This is because different package managers will have different ways of interpreting the forms passed to `:ensure`. I've updated the branch to re-add a normalizer that does nothing to its `args` list.

jwiegley:
Yes, but you deleted the normalizer, which will result in an error.

jwiegley:
I wonder if we're going about this the wrong way. Perhaps other package managers should just use a different keyword rather than `:ensure`?

raxod502:
@jwiegley That's fair—would it make more sense for the keyword to be called `:package`, then?

jwiegley:
I was thinking more `:straight`, for example. The only reason to make `:ensure` the front-end for any package manager, would be to allow transparent switching from one package manager to another, but I'm not convinced this is a use case we should optimize for -- since it makes the code just that much harder to test. If instead each package manager adds their own keyword to `use-package-keywords`, then they can do whatever they want, and advertise whatever sub-keywords they need to their users.

raxod502:
Yeah, I agree—but what I meant is that `straight.el` would have the keyword `:straight`, and `package.el` would have the keyword `:package`.

raxod502:
For backward compatibility, `:ensure` could just be an alias for `:package`. And then you get the "transparently switch between package managers" use case for free, because the user can just redefine `:ensure` as an alias for `:straight` instead, if they want.

jwiegley:
I suppose, if you're ok with accepting the same arguments that `:ensure` does.

raxod502:
Well, if we did as I propose, then there wouldn't be any concept of "the same arguments that `:ensure` does". But, yes, the idea of redefining `:ensure` to be an alias to `:straight` instead of `:package` would only work for those cases where `:straight` takes arguments of the same form as `:package`. That's why it wouldn't really be a supported use case. I just thought it would be a nice, elegant way to preserve backwards compatibility.

Besides, `:ensure` is kind of a weird name for a keyword. I feel like `:package` makes more sense, in addition to implicitly introducing a naming schema for other package managers.

There is the concern of deferred installation, which might be most elegant to implement in a package-manager-agnostic way. I have no objection, however, to making each package manager provide its own support for this, and it could also be argued that different package managers will probably be different enough that this would make more sense. (And of course *I* only care about deferred installation support for `straight.el` 😄)

raxod502:
In any case, I don't need any changes here to make `straight.el` work. What _would_ be nice is, as you mentioned in https://github.com/jwiegley/use-package/pull/552#issuecomment-349387607:

> `use-package-defaults` should probably be changed to use *pre-normalization* values, although right now it's using *post-normalization* values.

At the moment I kind of have to hack around in order to get my `:straight` keyword to default to the package name.

PR #592, opened 2017-12-16
Render the use-package-report table using Org

nickmccurdy:
I noticed that the "Gathering Statistics" instructions for rendering the Org
table could be automated for convenience. Now the table is reformatted
automatically, and you can still manually sort it. I'm not sorting it by default
to maintain the same order of use-package forms for easier troubleshooting.

### DIFF
--- a/use-package-core.el
+++ b/use-package-core.el
@@ -972,6 +972,9 @@ meaning:
                      (float-time (gethash :preface-secs hash 0))
                      (float-time (gethash :use-package-secs hash 0))))))
      use-package-statistics)
+    (goto-char (point-min))
+    (orgtbl-mode)
+    (orgtbl-ctrl-c-ctrl-c nil)
     (display-buffer (current-buffer))))
 
 (defun use-package-statistics-gather (keyword name after)
--- a/README.md
+++ b/README.md
@@ -736,8 +736,7 @@ initialization they've reached, and how much aggregate time they've spent
 (roughly), you can enable `use-package-compute-statistics` after loading
 `use-package` but before any `use-package` forms, and then run the command
 `M-x use-package-report` to see the results. The buffer displayed is an Org
-table for now, so just type `M-x orgtbl-mode` and then use `C-c C-c` to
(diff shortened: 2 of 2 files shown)


use-package-core.el
```diff
@@ -972,6 +972,9 @@ meaning:
                      (float-time (gethash :preface-secs hash 0))
                      (float-time (gethash :use-package-secs hash 0))))))
      use-package-statistics)
+    (goto-char (point-min))
+    (orgtbl-mode)
```
jwiegley: I originally had exactly this in the code, but wasn't sure if I wanted to depend on `orgtbl-mode` here, or if that's what users would want. Since Org comes with Emacs it's probably OK.

nickmccurdy: As long as Emacs 24.3 includes Org it should be fine (I think it does but if you're not sure we should double check). Even if it doesn't, we could mark org as a dependency. I would assume that as long as Org is built in the users would always want this, the table rendering was fairly quick and I found it hard to read before aligning it.

justbur:
As is the table is very difficult to read and not too useful.

My suggestion would be to either format it with org-table since that seems to be what is intended. You could always do

```elisp
(when (and (require 'org-table nil t)
           (fboundp 'org-table-align))
  (org-table-align))
```

to be safe about it. Doing it this way means it's unnecessary to go into org-mode. It's just using org's formatting function. The user could then enable org-mode if they wanted. This was my approach in #605 before I was aware of this PR.

Otherwise, it seems standard to use `tabulated-list-mode` to present a table with additional features like sorting.

dieggsy:
@justbur this PR suggests `orgtbl-mode`, not `org-mode`. Any reason not to prefer `orgtbl-mode`?

justbur:
@dieggsy my impression was that the minor mode `orgtbl-mode` is meant for situations where you want to edit org tables in other major modes, but since the buffer is in fundamental mode to start, I see no reason to not use plain `org-mode` here.

PR #596, opened 2017-12-17
Keep doc/ of the load-path

tarsius:
Without this `normal-top-level-add-subdirs-to-load-path`
would add this directory to the `load-path`.


kaushalmodi:
Where does this recursive path search happen? Is that `.nosearch` cookie file special for Melpa?

tarsius:
It happens when the Emacsmirror looks for all libraries that are part of a package.

It also currently happens when `borg` builds a package. But I will probably revert the change that I introduced in response to https://github.com/emacscollective/borg/issues/16, or turn it into an opt-in defect.

Issue #598, opened 2017-12-18
use-package conflict with global-company-mode


chenpoyang:
while editing use-package code:
<pre>
(use-package which-key
  :defer<---cursor here
</pre>
there is warnings in emacs:
<pre>
Error (use-package): Failed to parse package which-key: use-package: 
Unrecognized keyword: elisp--witness--lisp
</pre>

my company config is:
<pre>
"---------------------------------------------------------------------------"
;; company-mode
(require 'company)

;;(add-hook 'after-init-hook 'global-company-mode)
;;(setq company-global-modes '(not php-mode))
;;(setq company-global-modes '(not html-mode))

(global-company-mode 1)

(add-hook 'irony-mode-hook 'company-irony-setup-begin-commands)
(setq company-backends (delete 'company-semantic company-backends))

;; enable tab-completion with no delay
(setq company-idle-delay 0)
;;(define-key c-mode-map [(tab)] 'company-complete)
;;(define-key c++-mode-map [(tab)] 'company-complete)
</pre>

when i comment the following code:
<pre>
;; (global-company-mode 1)
</pre>

there is no warning while editing use-package code.

how to fix it ?

Luis-Henriquez-Perez:
This looks like the same problem I was talking about in at the end of #589. I am experiencing the same issue.

chenpoyang:
@Luis-Henriquez-Perez how to fix it?

chenpoyang:
debugger:
<pre>
Debugger entered--Lisp error: (error "use-package: Unrecognized keyword: elisp--witness--lisp")
  signal(error ("use-package: Unrecognized keyword: elisp--witness--lisp"))
  error("use-package: %s" "Unrecognized keyword: elisp--witness--lisp")
  use-package-normalize-plist(doom-themes (elisp--witness--lisp) nil use-package-merge-keys)
  use-package-normalize-keywords(doom-themes (elisp--witness--lisp))
[...]

PR #599, opened 2017-12-18
Handle automatic macro expansion by elisp-completion-at-point

justbur:
The function `elisp--local-variables` inserts this unbound variable into macro
forms to determine the locally bound variables for
`elisp-completion-at-point`. It ends up throwing a lot of errors since it's not
a keyword. Deleting it doesn't seem to break anything with completion, so ignore
it for now.

Related to https://github.com/abo-abo/hydra/commit/b12d37ac00d4f677b7684ed1af7d04d996167dbb

### DIFF
--- a/use-package-core.el
+++ b/use-package-core.el
@@ -559,6 +559,15 @@ extending any keys already present."
   (let* ((name-symbol (if (stringp name) (intern name) name))
          (name-string (symbol-name name-symbol)))
 
+    ;; The function `elisp--local-variables' inserts this unbound variable into
+    ;; macro forms to determine the locally bound variables for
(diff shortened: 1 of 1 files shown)


justbur:
I just noticed #598 which is what this addresses

justbur:
I'm more confident that this is the right approach. The purpose of inserting that variable is to discover locally bound variables, but there shouldn't be any new locally bound variables at the top level of the use-package declaration (right?). Note this change does not affect situations like this where there are local variables to learn about

```elisp
(use-package package
  :init
  (let ((a 1))
    elisp--witness--lisp))
```

PR #603, opened 2017-12-20
Fix sanity check for correct :custom-face format

### LINKED ISSUE #600: ":custom-face" not working with multiple faces
I'm trying to custom multiple faces but not working as expected. 

```emacs-lisp
(use-package spaceline-config
  :custom-face
  (powerline-inactive1 ((t (:background nil))))
  (powerline-inactive2 ((t (:background nil))))
  ;; (powerline-active1 ((t (:background nil :foreground nil)))) ; Uncommenting this line gets error
  )
```

This code with two faces works fine, but when I enable the third face line it shows error:

    Error (use-package): Failed to parse package spaceline-config: use-package: spaceline-config wants a (<symbol> <face-spec>) or list of these

Any help would be appreciated.

wbolster:
Instead of testing the lenght of each form passed to `:custom-face`,
the sanity check would test the number of forms passed to `:custom-face`,
causing it to fail when more than 2 face customisations are used.

Fixes #600.

### DIFF
--- a/use-package-core.el
+++ b/use-package-core.el
@@ -1351,7 +1351,7 @@ no keyword implies `:all'."
             (spec (nth 1 def)))
         (when (or (not face)
                   (not spec)
-                  (> (length arg) 2))
+                  (> (length def) 2))
           (use-package-error error-msg))))))
 
 (defun use-package-handler/:custom-face (name keyword args rest state)


wbolster:
dear 🎅 🤶,

all i want for christmas is custom font faces that actually work! 🎄

@wbolster  :trollface:

wbolster:
awesome, thanks. use-package is fantastic. :)

PR #605, opened 2017-12-20
Fix formatting of use-package-report table

justbur:
Use org-table-align to make it readable.


dieggsy:
Possible duplicate of #592

justbur:
whoops thanks

PR #617, opened 2018-01-27
[WIP] Lexical bindings

jabranham:
Using this reduces my Emacs load time a bit --- down from about 2.5 seconds to about 2.2 seconds. 

But some tests are failing at the moment. I think it's because lambdas get returned as closures, but I'll have to look later.

### DIFF
--- a/use-package-ensure.el
+++ b/use-package-ensure.el
@@ -76,15 +76,15 @@ The default value uses package.el to install the package."
 
 ;;;; :pin
 
-(defun use-package-normalize/:pin (name keyword args)
+(defun use-package-normalize/:pin (_name keyword args)
   (use-package-only-one (symbol-name keyword) args
-    #'(lambda (label arg)
-        (cond
-         ((stringp arg) arg)
-         ((use-package-non-nil-symbolp arg) (symbol-name arg))
-         (t
-          (use-package-error
-           ":pin wants an archive name (a string)"))))))
+    (lambda (_label arg)
+      (cond
+       ((stringp arg) arg)
+       ((use-package-non-nil-symbolp arg) (symbol-name arg))
+       (t
+        (use-package-error
+         ":pin wants an archive name (a string)"))))))
 
 (eval-when-compile
   (defvar package-pinned-packages)
@@ -116,7 +116,7 @@ manually updated package."
     (unless (bound-and-true-p package--initialized)
       (package-initialize t))))
 
-(defun use-package-handler/:pin (name keyword archive-name rest state)
+(defun use-package-handler/:pin (name _keyword archive-name rest state)
   (let ((body (use-package-process-keywords name rest state))
         (pin-form (if archive-name
                       `(use-package-pin-package ',(use-package-as-symbol name)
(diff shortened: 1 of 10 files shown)


npostavs:
> Why did you remove the sharp quotes from lambdas? That should hurt your performance, not help it.

It should have no effect on performance whatsoever, `lambda` is macro which adds the sharp quote, so the generated bytecode is the same.

jwiegley:
Then the change should have no meaning. I prefer the sharp quotes, so I'm happy to take the rest of the pull request, except for that commit.

jwiegley:
Please see http://endlessparentheses.com/get-in-the-habit-of-using-sharp-quote.html

jabranham:
I removed them so that they're properly fontified. Happy to take them out
though if you don't want them. Will do so later when I'm not on a phone.

jwiegley:
Thanks, @jabranham!  My only other question is: Have you signed over copyright with the FSF?  We intend `use-package` to be included in Emacs once I have a free moment, so I'm trying to ensure that all new contributions have been signed. Your help is much appreciated.

jabranham:
John, I'm confused now because that article says to never quote lamdas.

jabranham:
Yes, I've signed the copyright papers.

jabranham:
OK, I removed that commit.

use-package-ensure.el
```diff
@@ -76,15 +76,15 @@ The default value uses package.el to install the package."
 
 ;;;; :pin
 
-(defun use-package-normalize/:pin (name keyword args)
+(defun use-package-normalize/:pin (_name keyword args)
   (use-package-only-one (symbol-name keyword) args
-    #'(lambda (label arg)
```
jwiegley: Still an unneeded change here...

use-package-ensure.el
```diff
@@ -133,26 +133,26 @@ manually updated package."
 (defvar package-archive-contents)
 
 ;;;###autoload
-(defun use-package-normalize/:ensure (name keyword args)
+(defun use-package-normalize/:ensure (_name keyword args)
   (if (null args)
       (list t)
     (use-package-only-one (symbol-name keyword) args
-      #'(lambda (label arg)
```
jwiegley: And here...

jwiegley:
The article says that you never need to quote lambdas, but by doing so, you gain some diagnostics if you should misuse them.

npostavs:
> The article says that you never need to quote lambdas, but by doing so, you gain some diagnostics if you should misuse them.

No, the diagnostics only apply to sharp quoting symbols.

jabranham:
@jwiegley sorry, those should be fixed now.

jwiegley:
@jabranham I merged this, but then had to revert it.  It causes one of the tests to fail on Travis, and it breaks my configuration with errors like this:

```
Error (use-package): Failed to parse package alert: Symbol’s value as variable is void: arg
Error (use-package): Failed to parse package async: Symbol’s value as variable is void: arg
Error (use-package): Failed to parse package agda-input: Symbol’s value as variable is void: arg
Error (use-package): Failed to parse package chess: Symbol’s value as variable is void: arg
Error (use-package): Failed to parse package erc-yank: Symbol’s value as variable is void: arg
Error (use-package): Failed to parse package git-annex: Symbol’s value as variable is void: arg
Error (use-package): Failed to parse package git-undo: Symbol’s value as variable is void: arg
Error (use-package): Failed to parse package haskell-edit: Symbol’s value as variable is void: arg
Error (use-package): Failed to parse package initsplit: Symbol’s value as variable is void: arg
Error (use-package): Failed to parse package ledger-mode: Symbol’s value as variable is void: arg
Error (use-package): Failed to parse package regex-tool: Symbol’s value as variable is void: arg
Error (use-package): Failed to parse package smedl-mode: Symbol’s value as variable is void: arg
```

Until I'm sure this won't disrupt everyone, it'll have to wait until I have more time to examine.

jabranham:
Interesting. I'll look into it.

jabranham:
OK, @jwiegley I think I've figured it out. You're getting those errors on packages where you've used `:load-path` because you've adviced the function in your init file and the lexical bindings mess that up. You'd have to change it to:

```emacs-lisp
(defun load-path-handler-override (orig-func name _keyword arg rest state)
    (if (cl-some (apply-partially #'string-match load-path-reject-re) arg)
        (use-package-process-keywords name rest state)
      (let ((body (use-package-process-keywords name rest state)))
        (use-package-concat
         (mapcar #'(lambda (path)
                     `(eval-and-compile (add-to-list 'load-path ,path t)))
                 arg)
         body))))

  (advice-add 'use-package-handler/:load-path
              :around #'load-path-handler-override)
```

And then it should work.

But I'm not sure how to modify this so that it won't mess up anyone's config who has done something similar. Suggestions?

jwiegley:
Great analysis, thanks for doing that!

Since this is only a speed improvement, I think I'll hold off until after the 2.4 release. Otherwise, there will be a potential influx of tricky bug reports that I don't have time to field just now. :(

jabranham:
Sounds good! I'll rebase and open a new PR after the 2.4 release.

PR #624, opened 2018-02-06
Add use-package-ignore-unknown-keywords option

justbur:
This option allows use-package to skip over keywords it doesn't recognize.

The reason I want this option is that some keywords I use are tied to certain
packages being loaded, and I don't want a large chunk of my init file to fail if
any of those packages don't load.


jwiegley:
Hi Justin, this looks good! And thank you for signing your copyright papers already with the FSF. :)

PR #633, opened 2018-02-25
Support ":package" in use-package

matthewbauer:
bind-key supports the keyword ":package" but use-package does not know that. Adding this should be helpful. In the future, maybe we could automatically fill in with the use-package package, but that could be a breaking change.

### DIFF
--- a/use-package-bind-key.el
+++ b/use-package-bind-key.el
@@ -87,12 +87,14 @@ deferred until the prefix key sequence is pressed."
          ;;   :prefix STRING
          ;;   :filter SEXP
          ;;   :menu-name STRING
+         ;;   :package SYMBOL
          ((or (and (eq x :map) (symbolp (cadr arg)))
               (and (eq x :prefix) (stringp (cadr arg)))
               (and (eq x :prefix-map) (symbolp (cadr arg)))
               (and (eq x :prefix-docstring) (stringp (cadr arg)))
               (eq x :filter)
-              (and (eq x :menu-name) (stringp (cadr arg))))
+              (and (eq x :menu-name) (stringp (cadr arg)))
+              (and (eq x :package) (symbolp (cadr arg))))
           (setq args* (nconc args* (list x (cadr arg))))
           (setq arg (cddr arg)))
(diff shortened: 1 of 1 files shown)


nickmccurdy:
Does that just call a function with the package's name when the bind is pressed (like `C-s` for `swiper`, which is a function in the `swiper` package)?

matthewbauer:
So it should convert something like this:

```
(bind-keys
  :package view
  :map view-mode-map
  ("n" . next-line)
  ("p" . previous-line)
  ("j" . next-line)
  ("k" . previous-line)
  ("l" . forward-char)
  ("f" . forward-char)
  ("b" . backward-char)
  ("e" . end-of-line)
  ("a" . beginning-of-line))
```

into this:

```
(if
    (boundp 'view-mode-map)
    (progn
      (bind-key "n" #'next-line view-mode-map nil)
      (bind-key "p" #'previous-line view-mode-map nil)
      (bind-key "j" #'next-line view-mode-map nil)
      (bind-key "k" #'previous-line view-mode-map nil)
      (bind-key "l" #'forward-char view-mode-map nil)
      (bind-key "f" #'forward-char view-mode-map nil)
      (bind-key "b" #'backward-char view-mode-map nil)
      (bind-key "e" #'end-of-line view-mode-map nil)
      (bind-key "a" #'beginning-of-line view-mode-map nil))
  (eval-after-load 'view
    '(progn
       (bind-key "n" #'next-line view-mode-map nil)
       (bind-key "p" #'previous-line view-mode-map nil)
       (bind-key "j" #'next-line view-mode-map nil)
       (bind-key "k" #'previous-line view-mode-map nil)
       (bind-key "l" #'forward-char view-mode-map nil)
       (bind-key "f" #'forward-char view-mode-map nil)
       (bind-key "b" #'backward-char view-mode-map nil)
       (bind-key "e" #'end-of-line view-mode-map nil)
       (bind-key "a" #'beginning-of-line view-mode-map nil))))
```

jwiegley:
Looks good, although it's untested and undocumented for now.

Issue #641, opened 2018-03-08
Using tabulated-list for reporting statistics


DamienCassou:
The statistics buffer is in a fundamental mode with OrgTbl active. I find it a bit strange because Emacs features tabulated-list (used by, e.g., `list-processes`) which has the following strength:

- it is in special-mode with standard keybindings such as `q`, `h`, `<`...
- column titles are always visible even when scrolling down to the end of a long buffer
- column titles can be clicked to change sorting
- column titles show the sort status
- the buffer is less cluttered as columns are separated by spaces

IMO, OrgTbl is awesome for buffers mixing texts and tables which is not the case here.

I can implement it as I've used tabulated-list a lot.

/cc @nickmccurdy @justbur

justbur:
I think this is a good idea. If I remember right orgtbl was more of an easy upgrade than a permanent solution.

rchar01:
In my opinion tabulated-list for reporting statistics is a good idea.
Does `use-package-report` use some additional features of the OrgTbl package?

Now to load OrgTbl before `use-package-report`, I have `:commands (orgtbl-mode)` in my `use-package org` init part.

User can sort column e.g. in `list-buffers` buffer using `Buffer-menu-sort` command {cursor on specific column).

nickmccurdy:
Sounds cool, I didn't even know that mode could be reused. I'd be happy to review it.

jwiegley:
Nice, I'd actually prefer `tabulated-list` for this.

DamienCassou:
Please give me feedback on #642.

rchar01:
There are warnings after update to 20180308.1445 version:
```

In use-package-statistics-status:
use-package-core.el:970:27:Warning: reference to free variable ‘statistics’

In use-package-statistics-last-event:
use-package-core.el:979:44:Warning: reference to free variable ‘statistics’

In use-package-statistics-time:
use-package-core.el:986:40:Warning: reference to free variable ‘statistics’

In end of data:
use-package-core.el:1586:1:Warning: the function ‘hash-table-keys’ is not
    known to be defined.
```

After using the `M-x use-package-report`:
```
mapcar: Symbol’s function definition is void: hash-table-keys
```

PR #642, opened 2018-03-08
Use a tabulated-list to display package configuration statistics

DamienCassou:
Fix for #641


DamienCassou:
2018-03-08-204126_819x386_scrot

nickmccurdy:
Awesome, so the columns are in the same format but using tabulated-list instead of org?

DamienCassou:
> Awesome, so the columns are in the same format but using tabulated-list instead of org?

I don't understand the question. Indeed, the code uses tabulated-list to format the table instead of org. The columns are the same.

> Also please update the Gathering Statistics section of the readme.

good point. Done

PR #646, opened 2018-03-12
Revert "Fix report"

jwiegley:
Reverts jwiegley/use-package#645


jwiegley:
@DamienCassou This breaks the Travis build, claiming `subr-x` is not available.

Issue #644, opened 2018-03-10
use-package-report does not work


rchar01:
After using the `M-x use-package-report`:
```
mapcar: Symbol’s function definition is void: hash-table-keys
```

There are warnings after update to 20180308.1445 version:
```

In use-package-statistics-status:
use-package-core.el:970:27:Warning: reference to free variable ‘statistics’

In use-package-statistics-last-event:
use-package-core.el:979:44:Warning: reference to free variable ‘statistics’

In use-package-statistics-time:
use-package-core.el:986:40:Warning: reference to free variable ‘statistics’

In end of data:
use-package-core.el:1586:1:Warning: the function ‘hash-table-keys’ is not
    known to be defined.
```

DamienCassou:
That's my fault. I'm taking care of it (I got bitten by the fact that use-package-core doesn't use lexical binding). Which version of Emacs are you using?

DamienCassou:
Can you please try PR #645.

rchar01:
I use `GNU Emacs 25.3.1 (x86_64-unknown-cygwin, GTK+ Version 3.22.20)
 of 2017-09-11`.

rchar01:
`M-x use-package-report` works after adding `(require' subr-x) `.

DamienCassou:
I think the ticket should be reopened now that the fix has been reverted

PR #647, opened 2018-03-12
Fix report again after revert

DamienCassou:
This PR fixes both #650 and #644. The later has already been fixed by #645 but this was reverted in #646.


DamienCassou:
This is ready to be merged.

PR #651, opened 2018-03-14
Encourage installation via https instead of http

valignatev:
use-package is one of the most popular emacs packages, so it's better to encourage safer installation practices than plain `http://`. I hope I didn't miss anything.

P.S. Cheers @jwiegley, thanks for your awesome work!


jwiegley:
Thank you!

PR #657, opened 2018-03-20
Enable lexical binding

### LINKED ISSUE #648: Use lexical-binding
Why isn't use-package using lexical-binding for all files?

I volunteer to fix this if desired.

jabranham:
This supercedes #617 and closes #648

### DIFF
--- a/use-package-bind-key.el
+++ b/use-package-bind-key.el
@@ -1,4 +1,4 @@
-;;; use-package-bind-key.el --- Support for the :bind/:bind-keymap keywords
+;;; use-package-bind-key.el --- Support for the :bind/:bind-keymap keywords  -*- lexical-binding: t; -*-
 
 ;; Copyright (C) 2012-2017 John Wiegley
 
@@ -127,7 +127,7 @@ deferred until the prefix key sequence is pressed."
 
 ;;;###autoload
 (defun use-package-handler/:bind
-    (name keyword args rest state &optional bind-macro)
+    (name _keyword args rest state &optional bind-macro)
   (use-package-concat
    (use-package-process-keywords name rest state)
    `(,@(mapcar
@@ -148,7 +148,7 @@ deferred until the prefix key sequence is pressed."
 
 ;;;###autoload
 (defun use-package-handler/:bind-keymap
-    (name keyword args rest state &optional override)
+    (name _keyword args rest state &optional override)
   (use-package-concat
    (use-package-process-keywords name rest state)
    (mapcar
--- a/use-package-chords.el
+++ b/use-package-chords.el
@@ -1,4 +1,4 @@
-;;; use-package-chords.el --- key-chord keyword for use-package
+;;; use-package-chords.el --- key-chord keyword for use-package  -*- lexical-binding: t; -*-
 
 ;; Copyright (C) 2015-2017 Justin Talbott
 
--- a/use-package-core.el
+++ b/use-package-core.el
@@ -1,4 +1,4 @@
-;;; use-package-core.el --- A configuration macro for simplifying your .emacs
+;;; use-package-core.el --- A configuration macro for simplifying your .emacs  -*- lexical-binding: t; -*-
 
 ;; Copyright (C) 2012-2017 John Wiegley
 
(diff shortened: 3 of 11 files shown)


jabranham:
Looks like the test for `use-package-normalize-function` fails. I don't have time now, but will take a look later.

npostavs:
I think it's an error in the test, it looks to be making the exact same mistake as described in [Emacs Lisp Lambda Expressions Are Not Self-Evaluating](https://nullprogram.com/blog/2018/02/22/)

DamienCassou:
I've switched to your branch and will report if I find problems.

DamienCassou:
I've used this branch for 2 weeks now without problems. Can it be merged?

jwiegley:
I need to try it with my configuration first, since that failed spectacularly the last time this was tried. I'll let you know soon.

jabranham:
@jwiegley I did some digging last time and think I figure out what you need: https://github.com/jwiegley/use-package/pull/617#issuecomment-361023166

jwiegley:
I'm so glad you reminded me. Yes, that very likely is the problem.

I don't really use `:load-path` much anymore, since I configure Emacs using Nix these days and it manages the load path for me. So this is probably the right time to give this another try.

DamienCassou:
What is the status of this PR? I would like to get that in. It has been working for me for a month.

jabranham:
I'm not sure there's much to do on my end. I've been using it for a
while without any problems. Someone let me know if I'm mistaken and I
can take a look at whatever needs to be changed.

jwiegley:
I just need to give a try on my own config, it's on my list. Also, there's a test failing.

DamienCassou:
@jabranham could you please update your branch?

jabranham:
Rebased off master, thanks for the nudge @DamienCassou

jabranham:
@npostavs do you have a way to fix that test? I took a look at the blog post you shared and have tried various combinations of quoting without success so far.

npostavs:
I believe the quoting should be like this:
```elisp
(equal (use-package-normalize-function '(lambda () ...)) '(lambda () ...))
(equal (use-package-normalize-function ''(lambda () ...)) '(lambda () ...))
(equal (use-package-normalize-function '#'(lambda () ...)) '(lambda () ...))
```
But that still fails, I think because https://github.com/jwiegley/use-package/commit/9ce9c048a14967b4764872107c0ae9260febc44b#diff-8ef140df8b4408418fab781244480094L776 did not rewrite `use-package-normalize-function` correctly (the expressions above do all give `t` with the `pcase` version).

jabranham:
@npostavs They don't give `t` for me even with the pcase version. For example:

```
(equal (use-package-normalize-function '(lambda () ...)) (lambda () ...))
```

becomes

```
(equal (lambda nil \.\.\.) (closure (t) nil \.\.\.))
```

which evaluates to `nil`. I get the same thing with the current non-pcase version too.

npostavs:
> For example:
> 
>     (equal (use-package-normalize-function '(lambda () ...)) (lambda () ...))

You need to quote the right hand side `lambda` too.

jabranham:
D'oh! Of course.

jabranham:
So here's the reason for the test failing:

Doing
```emacs-lisp
(use-package-normalize-function ''(lambda () ...))
```

eventually tries to evaluate:

```emacs-lisp
(memq '(quote function) (car ''(lambda () ...))
```

which is a type error. The same thing happens if you try `(use-package-normalize-function '#'(lambda ...)`

jabranham:
Is it enough to just directly check if we're dealing with a quoted function?

```diff
From 894a3fe48c9251cf2103aada228f7b309f555a5a Mon Sep 17 00:00:00 2001
From: Alex Branham <branham@utexas.edu>
Date: Tue, 22 May 2018 12:10:21 -0500
Subject: [PATCH] Check for functions in use-package-normalize-function

---
 use-package-core.el | 2 ++
 1 file changed, 2 insertions(+)

diff --git a/use-package-core.el b/use-package-core.el
index 9f7828f..27232c8 100644
--- a/use-package-core.el
+++ b/use-package-core.el
@@ -919,6 +919,8 @@ If RECURSED is non-nil, recurse into sublists."
         ((and (consp v)
               (eq 'lambda (car v)))
          v)
+        ((functionp (nth 1 v))
+         (nth 1 v))
         ((and (listp v)
               (memq '(quote function) (car v))
               (eq 'lambda (car (cadr v))))
-- 
2.17.0

```

npostavs:
Actually, should that `(memq '(quote function) (car v))` be `(memq (car v) '(quote function))`?

jabranham:
Oh, right. And it's there, just a few lines up!

jabranham:
Rebased off master

jwiegley:
This is working great for me now, thanks. I just had to fix a bug in my `load-path-handler-override`, which I'm surprised was never causing an issue before.

PR #658, opened 2018-03-24
New style of auto-deferral for chords

killdash9:
new style of auto-deferral for chords


killdash9:
I'll explain what it means first, and then you can decide if you still want something added to the manual.  I'm not introducing any new functionality here, I'm fixing a bug.  The bug is that the plugin has stopped registering the commands that are bound to the chords as autoloaded functions.  This meant that pressing a key chord would result in an error about the symbol not being a command, unless the command was explicitly registered with the `:commands` keyword.

This bug began when you comitted https://github.com/jwiegley/use-package/commit/b0e53b4238f027350e16aeef712a610c9d2c9961 which "Changes to the way auto-deferral is indicated".  

The old way of doing this was for the plugin to append to the `:commands` part of the data structure inside of the `use-package-handler:/foo` function.  The new way is to define a `use-package-autoloads/:foo` So my patch is simply to bring it up to date with the new way of doing things.

killdash9:
No I haven't done that.  I'll look into it.

killdash9:
@jwiegley  I have completed the FSF copyright assignment.  Do you need me to send you the assignment pdf and signature files as proof?  I'm anxious to get this patch applied.

PR #663, opened 2018-03-29
Add info about auto-package-update to the README

asmeurer:
When I first read the README, I assumed that :ensure kept packages up-to-date.
It is very unclear about this. Having discovered that it doesn't, I found
auto-package-update, which seems to do this job just fine.


nickmccurdy:
Thanks, wish I heard about this sooner. 😄

PR #673, opened 2018-04-25
[Fix #661] ensure-system-package: honor system-packages customizations

### LINKED ISSUE #661: :ensure-system-package ignores system-packages-noconfirm

akhramov:
ensure-system-package doesn't honor system-packages
customizations (#661), because system-packages didn't provide an API
to retrieve the shell command to be executed.

This change makes use of the new system-packages'
`system-package-get-command` function and therefore fixes the issue.

See also jabranham/system-packages#25


akhramov:
@tarsius thank you for pointing this out! I've created #683 to address this particular issue.

tarsius:
Thanks!

PR #686, opened 2018-06-04
Correct prefix keys after use-package-autoload-keymap

justbur:
By adding events of the form (cons t event) to unread-command-events, Emacs
correctly updates this-command-keys for the subsequent use of the prefix.

Before this change, packages like my which-key were not getting the correct
prefix after use-package-autoload-keymap loaded the relevant package.

See https://github.com/justbur/emacs-which-key/issues/192

### DIFF
--- a/use-package-bind-key.el
+++ b/use-package-bind-key.el
@@ -61,7 +61,8 @@ deferred until the prefix key sequence is pressed."
               (bind-key* key keymap)
             (bind-key key keymap))
           (setq unread-command-events
-                (listify-key-sequence kv)))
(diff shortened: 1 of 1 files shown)


mambolevis:
Hi, I'am waiting and ready for making tests if required.

andschwa:
I tested and this fixed the issue for me too. I can now use `:bind-keymap ("C-c &" . yas-keymap)` and the `which-key` guide is correctly filled in with the `yas-keymap` after the package is loaded (and previously it just showed the guide for `C-c`).

andschwa:
@jwiegley This PR has been working great, just FYI 😄

PR #689, opened 2018-06-12
Update debug info in readme

### LINKED ISSUE #626: Show expanded code
Hello,

Is there any way to (manually) display the expanded code of a `use-package` expression ?

Commit 418e90c39a removed the `use-package-debug` option, and IIRC I could do `C-x C-e` and it would show me the expanded code in a buffer.

As I understand it the _new_ `debug` choice of `use-package-verbose` do the same thing but only when an error append.
A workaround is to explicitly throw an error, e.g `(use-package diminish :config (error "expand-me"))`.

Maybe there could be a `use-package-eval` function that would evaluate and show the expanded code at point ?

Also there still is a mention of this behavior here https://github.com/jwiegley/use-package/blame/master/README.md#L893
I can correct it if needed and I'm not mistaken on what's happening.

Thanks for this awesome package.

andschwa:
This resolves #626.

### DIFF
--- a/README.md
+++ b/README.md
@@ -927,9 +927,8 @@ body, so that only the minimum code necessary is emitted as the result of a
 
 After the keyword has been inserted into `use-package-keywords`, and a
 normalizer and a handler defined, you can now test it by seeing how usages of
-the keyword will expand.  For this, temporarily set `use-package-debug` to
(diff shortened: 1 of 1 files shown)


andschwa:
Thanks!

PR #697, opened 2018-06-29
Place :chords at beginning of use-package-keywords

killdash9:
When it is at the end, the keys don't get bound until after the package has been loaded, because it comes after :defer, which is bad.  Placing it at the beginning allows the key bindings to be mapped to auto-loaded functions before the package is loaded, so that a key chord press triggers the package load.

### DIFF
--- a/use-package-chords.el
+++ b/use-package-chords.el
@@ -37,7 +37,7 @@
    `(,(macroexpand
        `(bind-chords :package ,name ,@arg)))))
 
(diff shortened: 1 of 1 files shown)


waymondo:
I can confirm the issue is real and this PR fixes it.

waymondo:
@jwiegley there isn't much to review here, unless you have any concerns about the keyword handling ordering. I feel comfortable merging or will don you the honors.

Issue #698, opened 2018-06-29
:ensure-system-package error when package manager is nil


markcol:
If a `use-package` declaration contains an `:ensure-system-package clause`, but `system-packages-package-manager` is nil, the package will generate an error and not be loaded. This is an issue when the configuration is used on multiple platform (e.g.: Unix, Windows) where one of the platforms does not have a supported package manager.

The following `use-package` declaration:

```
(use-package graphviz-dot-mode
  :mode "\\.dot\\'"
  :ensure-system-package graphviz)
```
generates the following error when no package manager is available:

```
Error (use-package): Failed to parse package graphviz-dot-mode: install not supported in nil
```

A more user-friendly solution would be to continue to load the package, but generate a warning that the requested system package cannot be verified. This would allow the same configuration to be used in both environments without issues, though the system package dependencies would have to be satisfied manually for any environment that does not have a supported package manager.

PR #699, opened 2018-06-29
Issue warning when no package manager available.

markcol:
When :ensure-system-package is encountered, but
system-packages-package-manager is nil, a warning will be issued that
the package cannot be verified rather than causing an error that
results in the enclosing use-package declaration to fail.

Fixes: https://github.com/jwiegley/use-package/issues/698


jwiegley:
It looks like this branch no longer applies?

waymondo:
Yes, I think this PR is safe to close. `system-packages-install` is now called directly. I would expect any warning behavior that a package manger is not available to come from the `system-packages` packages itself.

Issue #717, opened 2018-09-02
Check if package is installed


holocronweaver:
Does `use-package` provide a package backend agnostic way to check if a package is installed?

For example, while `package` is  a popular backend, I use `straight.el` with `use-package`. If I want to write a config that is independent of `use-package` backend, how can I check whether packages are installed?

Or is doing such a check before installing performance-wise unnecessary with `use-package`?

For example, with `package` I used to do:

```
(unless (package-installed-p package)
  (package-install package)))
```

As a related followup, is there a backend agnostic way to get a list of installed packages?

slippycheeze:
I'm not the upstream, but I can answer your question: `use-package` implements the specific logic you described there in the default, package.el based `use-package-ensure-elpa`: [use-package-ensure.el line 155](https://github.com/jwiegley/use-package/blob/master/use-package-ensure.el#L155)

The logic, sensibly, is entirely within that function, so that when, eg, `straight.el` overrides it with their own implementation (and thus takes over the `:ensure` keyword entirely), it has to implement the same check.

Which, of course, would have packages installed, but also, `package-installed-p` returning `nil` as there was no corresponding ELPA package present on the system.  

There is no broader abstraction over package systems present.

holocronweaver:
I think the key takeaway is that calling `use-package` multiple times on the same package probably won't have much if any performance impact. While it would be nice to have `use-package` provide more convenience wrappers which backends can override, I don't think it's necessary in my use case.

PR #724, opened 2018-09-21
Add comparison to package managers

bradrn:
In many tutorials, one of the first features of `use-package` mentioned after its use is introduced is its ability to call package managers. Here's three examples from three different tutorials:

> With it [`use-package`], I can automatically download and install other packages from the emacs package databases when it detects they're missing.

> ```
> (use-package smex
>  :ensure t
>  :bind (("M-x" . smex))
>  :config (smex-initialize))
> ```
> And now Emacs ensures the package is installed [...]

> ```
> (use-package general :ensure t)
> ```
> It should check for the general package and make sure it is accessible. If not, the :ensure t part of the previous chunks tells use-package to download it [...]

While this focus on downloading packages is understandable, as it is one of the most convenient pieces of functionality provided by `use-package`, it could be confusing to newcomers such as me who could confuse `use-package` with a package manager. I do know that in my case, this caused a horrible amount of mental confusion, which was only clarified once I saw #717.

To reduce this confusion I have added a small note to the top of `use-package` clarifying this in the hope that it could help other people who were confused like me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,11 @@ created it because I have over 80 packages that I use in Emacs, and things
 were getting difficult to manage.  Yet with this utility my total load time is
 around 2 seconds, with no loss of functionality!
 
+Note that `use-package` is **not** a package manager! Although `use-package`
+does have the useful capability to interface with package managers (see
+[below](#package-installation)), its primary purpose is for the configuration
+and loading of packages.
+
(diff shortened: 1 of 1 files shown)


tarsius:
I've often seen users being confused about this and think that adding this clarification would help a lot.

PR #733, opened 2018-11-10
Silence byte-compiler on Emacs 25

tarsius:
Emacs 25 defined a global variable `features`, which triggers a
warning "Lexical argument shadows the dynamic variable features".
That's not `use-package`'s fault, but we should suppress the warning
anyway, so that there is no additional noise that would cause us to
potentially overlook warnings that absolutely have to be addressed.

### DIFF
--- a/use-package-core.el
+++ b/use-package-core.el
@@ -1315,41 +1315,41 @@ meaning:
       args
     (list args)))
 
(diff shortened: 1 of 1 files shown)


tarsius:
That was quick. Thanks!

PR #744, opened 2019-02-10
use `use-package-as-one` for normalizing `:ensure-system-package`

waymondo:
This makes the optional syntax consistent with other `use-package`
keywords. Now both of these are valid:

```
(use-package format-all
  :ensure-system-package
  (prettier . "npm i -g prettier")
  (rufo . "gem install rufo"))

(use-package format-all
  :ensure-system-package
  ((prettier . "npm i -g prettier")
   (rufo . "gem install rufo")))
```


waymondo:
Hi @jwiegley, I noticed you were doing lightweight PR merging. Want to grab this as well?

PR #749, opened 2019-03-03
Adding support for tags

seblemaguer:
Hello,

as I have multiple computers and I like classifying packages, I added some tag support. It was also the first time I was playing with extending use-package so I suspect I did a lot of stuff wrong so I do this pull request to get also some feedbacks.

### DIFF
--- a/use-package-tags.el
+++ b/use-package-tags.el
@@ -1,0 +1,75 @@
+;;; use-package-tags.el ---   -*- lexical-binding: t; coding: utf-8 -*-
+
+;; Copyright (C)  2019 Sébastien Le Maguer
+;;
+
+;; Author: Sébastien Le Maguer <lemagues@tcd.ie>
+;; Maintainer: Sébastien Le Maguer <lemagues@tcd.ie>
+;; Created: 03 Mar 2019
+;; Modified: 03 Mar 2019
+;; Version: 1.0
+;; Package-Requires: ((use-package "2.4") (dash "2.15.0"))
+;; Keywords: dotemacs startup package tags
+;; URL: https://github.com/jwiegley/use-package
+
+;; use-package-tags is free software; you can redistribute it and/or modify it
+;; under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 3, or (at your option)
+;; any later version.
+;;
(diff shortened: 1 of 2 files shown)


use-package-tags.el
```diff
@@ -0,0 +1,75 @@
...
+  package and the list of tags associated the hostname of the
+  current machine is not empty.")
+
+;; 0. Define some helper
+(defun use-package-tags-validate-tags (tags)
+  "Helper to see if the given TAGS may lead to an
+ installation. Returns t if TAGS is nil or the list of tags
+ associated to the current machine hostname is nil. Else, it
+ returns the intersection between the two lists of tags."
+  (let* ((list-tags-cur-host (alist-get (system-name) use-package-tags-enabled nil nil 'string=)))
```
a13: afaik, alist-get appeared only in emacs 25.1, so you have to add the dependency to Package-Requires section

use-package-tags.el
```diff
@@ -0,0 +1,75 @@
...
+
+;;
+
+;;; Code:
+
+(require 'use-package-core)
+(require 'dash)
+
+;; Define some machine name identifiers
+(defvar use-package-tags-enabled '()
```
a13: why not `defcustom` ?

jwiegley:
The only reason I hesitate to add this, is that you could already achieve this using `:if`:

```
(defconst this-machine ... some logic to determine a label for this machine ...)

(defun machine-tags (machine) ... a mapping of machines to tags ...)

(defun has-tag (tag) ... returns t if the current machine supports this tag ...)

(use-package foo
  :if (has-tag 'work)
```

I don't know if there need to be another "layer" for something that can be done with Lisp...

seblemaguer:
Thanks a lot for the comment. I propose that I start do that in a package aside to begin and to see what can be the development of this and if it proves useful, I will open a new pull request. Is it ok?

@a13 thanks for the comments I update that on my side

jwiegley:
@seblemaguer If you find that built-in tags support offers better utility or expressiveness, I'll be happy to merge it.

seblemaguer:
I close this pull request for now. I keep you in touch about updates if necessary

Issue #709, opened 2018-08-04
defer accepts an optional float argument?


FirstLoveLife:
[Here](https://github.com/jwiegley/use-package#defer-now-accepts-an-optional-integer-argument) says it only accepts an integer, but when testing evil-----:defer 0.5, it also works as expected.

PR #750, opened 2019-03-04
Update the README for defer

rememberYou:
Some users still think that `:defer` only accepts an integer as an argument, while it can take a float (**SEE:** https://github.com/jwiegley/use-package/issues/709).

To further avoid this confusion, the `README.md` file has been updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1037,7 +1037,7 @@ this case, but that's a subtlety I'd rather avoid.
 ## :defer now accepts an optional integer argument
 
 `:defer [N]` causes the package to be loaded -- if it has not already been --
-after `N` seconds of idle time.
+after `N` seconds of idle time, where `N` is a float number.
 
 ``` elisp
 (use-package back-button


thomasf:
Well, it's actually just a number which can be a integer or a float.

rememberYou:
You're right, but maybe we can say it explicitly? It should not be forgotten that the [README](url) must be understood for a user who does not have knowledge of ELisp.

thblt:
Your changes are incorrect, the value is not a float but any number.  So the line you modified should be left as it was; instead you could modify line  1037 from

     ## :defer now accepts an optional integer argument 

to

     ## :defer now accepts an optional numeric argument

rememberYou:
@thblt thank you, I fixed the change [here](https://github.com/jwiegley/use-package/pull/754), my apologies for the double pull request.

thblt:
@rememberYou fwiw you could have `--amend`

PR #778, opened 2019-06-20
Update bind-chords to use eval-after-load when keymaps are declared but not bound

### LINKED ISSUE #777: :chords local keymap doesn't work?
```
(use-package org
  ....

  ;; Uncommenting this leads to an error (void-variable org-mode-map)
  ;; :chords
  ;; (:map org-mode-map
  ;;       ("oo" . hydra-org-template/body))

   ;; This works
  :bind
  (:map org-mode-map
        ("C-<" . hydra-org-template/body))
        )
```
I'm having a bit of trouble binding a chord to a local keymap - I'm probably doing something wrong, help would be appreciated. I'm using straight.el if that's relevant, and I'm able to use `:chords`  for global key chords without any problems.

waymondo:
Also improve :chord keyword syntax processing to more closely mimic bind-keys since the same binding normalizer function is used.

Also add tests for use-package-chords to cover these test cases.

Update homepage url and version for both `bind-chord` and `use-package-chords`.

Closes #777.

### DIFF
--- a/bind-chord.el
+++ b/bind-chord.el
@@ -1,11 +1,11 @@
 ;;; bind-chord.el --- key-chord binding helper for use-package-chords
 
-;; Copyright (C) 2015-2017 Justin Talbott
+;; Copyright (C) 2015-2019 Justin Talbott
 
 ;; Author: Justin Talbott <justin@waymondo.com>
 ;; Keywords: convenience, tools, extensions
-;; URL: https://github.com/waymondo/use-package-chords
-;; Version: 0.2
+;; URL: https://github.com/jwiegley/use-package
+;; Version: 0.2.1
 ;; Package-Requires: ((bind-key "1.0") (key-chord "0.6"))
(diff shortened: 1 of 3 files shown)


waymondo:
Hi @jwiegley, could I ask for you to review my PRs here or would you like me to remove `use-package-chords` and `use-package-ensure-system-packages` from your repo? I like the idea of the extensions living here, but I'm finding it hard to support them without the ability to merge fixes, tests, etc.

jwiegley:
@waymondo I've been a bit absent from GitHub recently; I'm fine if you want to move the code out, and would completely understand. I'll keep more observant of these PRs in future.

waymondo:
Thanks for merging these. I've been using a fork of `use-package` with these changes on them since 2019, so it will be nice to rebase back to master. I'll leave the `use-package` extensions in this repo for now, since I haven't felt the need to modify them since.

PR #783, opened 2019-07-12
Use `require', not `load', when byte-compiling

raxod502:
I would like to propose the following change to `use-package`: when byte-compiling, use `require` to load packages, rather than `load`. This means that packages are not loaded twice, which is slow and can sometimes break things if the user's configuration overrides something that was defined in a package.

### DIFF
--- a/use-package-core.el
+++ b/use-package-core.el
@@ -651,7 +651,8 @@ extending any keys already present."
                     ,(when (eq use-package-verbose 'debug)
                        `(message ,(format "Compiling package %s" name-string)))
                     ,(unless (plist-get args :no-require)
-                       `(require ',name-symbol)))))))))
+                       `(unless (featurep ',name-symbol)
+                          (load ,name-string nil t))))))))))
 
     ;; Certain keywords imply :defer, if :demand was not specified.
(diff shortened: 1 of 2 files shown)


raxod502:
The CI build appears to be failing for a reason unrelated to my change.

jwiegley:
What about files that don't provide a feature? This could be a destabilizing change.

raxod502:
That is a good point. What if we stick with `load`, but refrain from invoking it when a feature by that name has been provided already?

jwiegley:
@raxod502 Sounds like a good compromise to me!

raxod502:
Alright, I've changed the approach and updated the tests so they pass locally. Does this look good to merge?

jwiegley:
Travis seems borked, so I'll just ask if you've run all the testing on your side and run this using your own configuration?

raxod502:
Yep, I've run `make test` locally and it works with [Radian](https://github.com/raxod502/radian). I might be able to fix Travis for you, by porting the Docker workflow I set up for [Apheleia](https://github.com/raxod502/apheleia) from Semaphore to Travis.

PR #799, opened 2019-10-30
Add optional docstring support to `use-package`.

cireu:
Add docstring support to `use-package`

```elisp
(use-package xxx
   "Description"
   ...)
```

I think it's useful for user to describe the usage of a  `use-package` block.

### DIFF
--- a/use-package-core.el
+++ b/use-package-core.el
@@ -1515,6 +1515,7 @@ For full documentation, please see the README file that came with
 this file.  Usage:
 
   (use-package package-name
+     [docstring]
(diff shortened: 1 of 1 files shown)


jwiegley:
Since there's nothing that would use that string, use a Lisp comment:

```lisp
(use-package xxx
   ;; Description
   ...)
```

cireu:
thanks, I'll use custom `:doc` handler to use traditional string style doc :)

jwiegley:
@cireu Sounds like a good compromise! :)

PR #820, opened 2020-03-05
Clarify the documentation for :after

### LINKED ISSUE #705: :bind defeats :after
I had this in my init file:

```lisp
(use-package bicycle :after outline)
```

This expands to something involving `(require 'bicycle)`, but only after `outline` is loaded.

But then I added my personal key bindings:

```lisp
(use-package bicycle
  :after outline
  :bind (:map outline-minor-mode-map
              ([C-tab] . bicycle-cycle)
              ([S-tab] . bicycle-cycle-global)))
```

And now the expansion involves `autoload` and the `require` is gone. It seems to me that `:after` should be enough to cause the package to be loaded `:after` the other package, regardless of what other keywords are involved.

I was able to force the `require` to reappear by adding `:demand`, which resulted in exactly the behavior I would have expected *without* that addition: `bicycle`'s loading was once more deferred until after `outline` was loaded. However, that does not match the documented behavior of `:demand`, which claims that it "prevent[s] deferred loading in all cases". In this case it does not prevent deferred loading, it enables it.

jwiegley:
Fixes #705

### DIFF
--- a/use-package-core.el
+++ b/use-package-core.el
@@ -1548,9 +1548,11 @@ this file.  Usage:
                  `:magic-fallback', or `:interpreter'.  This can be an integer,
                  to force loading after N seconds of idle time, if the package
                  has not already been loaded.
-:after           Defer loading of a package until after any of the named
-                 features are loaded.
-:demand          Prevent deferred loading in all cases.
+:after           Delay the use-package declaration until after the named modules
+                 have loaded. Once load, it will be as though the use-package
+                 declaration (without `:after') had been seen at that moment.
+:demand          Prevent the automatic deferred loading introduced by constructs
+                 such as `:bind' (see `:defer' for the complete list).
 
 :if EXPR         Initialize and load only if EXPR evaluates to a non-nil value.
 :disabled        The package is ignored completely if this keyword is present.


use-package-core.el
```diff
@@ -1548,9 +1548,11 @@ this file.  Usage:
                  `:magic-fallback', or `:interpreter'.  This can be an integer,
                  to force loading after N seconds of idle time, if the package
                  has not already been loaded.
-:after           Defer loading of a package until after any of the named
-                 features are loaded.
-:demand          Prevent deferred loading in all cases.
+:after           Delay the use-package declaration until after the named modules
+                 have loaded. Once load, it will be as though the use-package
```
progfolio: s/load/loaded

tarsius:
It's still not great and you should probably talk about it some more in the readme.

I would change "the use-package declaration" to "the effect of the rest of the declaration" and later just "the declaration".

More importantly though `:after` should not be placed between `:defer` and `:demand`, because that location implies the misconception we try to prevent.  It could be placed just below (with an empty line in between, or maybe even as the very first item.

``` diff
diff --git a/use-package-core.el b/use-package-core.el
@@ -1548,12 +1548,14 @@ (defmacro use-package (name &rest args)
                  `:magic-fallback', or `:interpreter'.  This can be an integer,
                  to force loading after N seconds of idle time, if the package
                  has not already been loaded.
-:after           Delay the use-package declaration until after the named modules
-                 have loaded. Once load, it will be as though the use-package
-                 declaration (without `:after') had been seen at that moment.
 :demand          Prevent the automatic deferred loading introduced by constructs
                  such as `:bind' (see `:defer' for the complete list).
 
+:after           Delay the effect of the rest of the declaration until after
+                 the named modules have loaded.  Once they have been loaded,
+                 it will be as though the declaration (without `:after') had
+                 been seen at that moment.
+
 :if EXPR         Initialize and load only if EXPR evaluates to a non-nil value.
 :disabled        The package is ignored completely if this keyword is present.
 :defines         Declare certain variables to silence the byte-compiler.
```

Maybe also replace the second sentence.

```
:after           Delay the effect of the rest of the declaration until after
                 the named modules have loaded.  Before they have been loaded
                 no other keyword is of any relevance at all and once they
                 have been loaded it is as if `:after' was not specified.
```

tarsius:
Please also double-check whether `:defer` mentions all the relevant keywords, I think it doesn't.

tarsius:
... on the other hand -- it is better what we have now.

jwiegley:
I'll merge this, but create an issue to pursue your suggestion.

tarsius:
Thanks!

PR #830, opened 2020-04-15
Allow multiple keymaps in :map argument

fishyfriend:
This PR augments bind-keys to support passing a list of keymaps as the `:map` argument.

When one wants to bind the same key/command pair in multiple keymaps, the current options have some drawbacks:

```elisp
;; Option 1: Code duplication, problematic when duplicated keybindings are numerous
(use-package foopkg
   :bind (:map bar-mode-map
          ("C-c x" . foocmd1)
          ("C-c y" . foocmd2)
          :map baz-mode-map
          ("C-c x" . foocmd1)
          ("C-c y" . foocmd2)))

;; Option 2: Use a loop, sacrificing the convenience and readability of :bind
(use-package foopkg
  :commands (foocmd1 foocmd2)
  :init
  (dolist (map (list bar-mode-map baz-mode-map))
    (bind-keys :map map
               ("C-c x" . foocmd1)
               ("C-c y" . foocmd2))))
```

With the new way, it is possible to avoid duplication and still use `:bind`:

```elisp
(use-package foopkg
  :bind (:map (bar-mode-map baz-mode-map)
        ("C-c x" . foocmd1)
        ("C-c y" . foocmd2)))
```

Additionally this PR clarifies the documentation of `bind-chords` to reflect that this style is already available for that command.

### DIFF
--- a/bind-key.el
+++ b/bind-key.el
@@ -248,12 +248,12 @@ In contrast to `define-key', this function removes the binding from the keymap."
   "Similar to `bind-key', but overrides any mode-specific bindings."
   `(bind-key ,key-name ,command override-global-map ,predicate))
 
-(defun bind-keys-form (args keymap)
+(defun bind-keys-form (args &rest keymaps)
   "Bind multiple keys at once.
 
 Accepts keyword arguments:
 :map MAP               - a keymap into which the keybindings should be
-                         added
+                         added, or a list of such keymaps
 :prefix KEY            - prefix key for these bindings
 :prefix-map MAP        - name of the prefix map that should be created
                          for these bindings
@@ -276,7 +276,7 @@ Accepts keyword arguments:
 
 The rest of the arguments are conses of keybinding string and a
 function symbol (unquoted)."
-  (let (map
+  (let (maps
         prefix-doc
         prefix-map
         prefix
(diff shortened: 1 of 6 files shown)


use-package-tests.el
```diff
@@ -1951,15 +1951,92 @@
...
+(ert-deftest bind-key/:repeat-map-2 ()
+  (match-expansion
+   (bind-keys :map m ("x" . cmd1) :repeat-map rm ("y" . cmd2))
+   `(progn
+      (bind-key "x" #'cmd1 m nil)
+      (defvar rm (make-sparse-keymap))
+      (put #'cmd2 'repeat-map 'rm)
+      (bind-key "y" #'cmd2 rm nil))))
+
+(ert-deftest bind-key/:repeat-map-3 ()
```
fishyfriend, line 2030: The expansion in this test is not equivalent to the prior test as might be expected. The `repeat-map` property is applied to bound commands even after an intervening `:map` directive. This seems fishy to me but is existing behavior (the tests both pass independently of the changes in this PR).

bind-key.el
```diff
@@ -293,20 +293,17 @@ function symbol (unquoted)."
       (while (and cont args)
         (if (cond ((and (eq :map (car args))
                         (not prefix-map))
-                   (setq map (cadr args)))
+                   (let ((arg (cadr args)))
+                     (setq maps (if (listp arg) arg (list arg)))))
                   ((eq :prefix-docstring (car args))
                    (setq prefix-doc (cadr args)))
-                  ((and (eq :prefix-map (car args))
```
fishyfriend, line 299: I've removed this condition `(not (memq map '(global-map override-global-map)))` as, without understanding its purpose, I was not sure how to modify it to support the new list form of `:map`. Is there a reason why one could not use `:prefix-map` with the global maps?

bind-key.el
```diff
@@ -293,20 +293,17 @@ function symbol (unquoted)."
...
-                        (not (memq map '(global-map
-                                         override-global-map))))
+                  ((eq :prefix-map (car args))
                    (setq prefix-map (cadr args)))
                   ((eq :repeat-docstring (car args))
                    (setq repeat-doc (cadr args)))
-                  ((and (eq :repeat-map (car args))
```
fishyfriend, line 305: Same rationale & question as above.

fishyfriend:
Hi @jwiegley, thanks. I've revised this and resolved the conflicts. I made one assumption that could use a second-guess, please see [this comment](https://github.com/jwiegley/use-package/pull/830/files/d3e92396e1a0df05c07906b6bf55d4d7e3746562#r983269136). Other than that, it should be good to go. Cheers, Jacob. (cc: @Hugo-Heagren)

jwiegley:
I was unable to update this PR directly, so I've opened another PR with the same changes @fishyfriend.

PR #834, opened 2020-04-26
Fix typo

tarsius:


tarsius:
Every few months I run the spell checker on all the packages I have installed and fix the new typos.
In this case I ended up fixing the same typos again, so I thought I would send this friendly ping.

Ping!

tarsius:
Pong?

Regarding the failed test:

> /tmp/emacs/bin/emacs: error while loading shared libraries: libgnutls.so.30: cannot open shared object file: No such file or directory

I don't think my typo fix in the readme did cause that. :grin:

Issue #282, opened 2015-11-11
Getting assignment forms before submission into Emacs


jwiegley:
- [x] ~@BernardHurley~
- [x] @DamienCassou (Damien Cassou)
- [x] @Fuco1 (Matus Goljer)
- [x] ~@MaikuMori (Miks Kalnins)~
- [x] @PhilHudson (Phil Hudson)
- [x] @Silex (Philippe Vaucher)
- [x] @YoungFrog (Nicolas Richard)
- [x] @aethanyc (Ting-Yu Lin)
- [x] @alezost (Alex Kost)
- [ ] @andschwa (Andy Schwartzmeyer)
- [x] @aspiers (Adam Spiers)
- [x] @basil-conto (Basil L. Contovounesios)
- [x] ~@bdd (Berk D. Demir)~
- [x] @canatella (Damien Merenne)
- [x] ~@carldotac (Carl Lieberman)~
- [x] ~@cmpitg (Ha-Duong Nguyen)~
- [x] @dabrahams (Dave Abrahams)
- [x] @dudebout (Nicolas Dudebout)
- [x] ~@edvorg (Eduard Knyshov)~
- [x] @ffevotte (François Févotte)
- [x] ~@hartzell (George Hartzell)~
- [x] @jabranham (Alex Branham)
- [x] @joewreschnig (Joe Wreschnig)
- [x] @justbur (Justin Burkett)
- [x] @kaushalmodi (Kaushal Modi)
- [x] @killdash9 (Russell Black)
- [x] @kovrik
- [x] ~@marcwebbie (Marc Webbie)~
- [x] ~@matklad (Aleksey Kladov)~
- [x] ~@mcandre (Andrew Pennebaker)~
- [x] ~@milkypostman (Donald Curtis)~
- [x] @nickmccurdy (Nick McCurdy)
- [x] @npostavs (Noam Postavsky)
- [x] @peterhoeg (Peter Hoeg)
- [x] @phillord (Phil Lord)
- [x] @phst (Philipp Stephani)
- [x] @purcell (Steve Purcell)
- [x] @raxod502 (Radon Rosborough)
- [x] ~@rememberYou (Terencio Agozzino)~
- [x] ~@syl20bnr (Sylvain Benner)~
- [x] @tarsius (Jonas Bernoulli)
- [x] ~@thiagowfx (Thiago Perrotta)~
- [x] @thomasf (Thomas Frössman)
- [x] @vermiculus (Sean Allred)
[...]

PR #840, opened 2020-05-09
Call hooks even if no config

### LINKED ISSUE #785: use-package-inject-hooks {pre,post}-config not run if :config block not present
This is very similar to https://github.com/jwiegley/use-package/issues/250 -- but that was years ago, and this issue is specifically about config, not init.

The use-package docs for `use-package-inject-hooks` say:

    NOTE: These hooks are run even if the user does not specify an
    `:init' or `:config' block, and they will happen at the regular
    time when initialization and configuration would have been
    performed.

That seems to be untrue for `:config`.  use-package master has this code in `use-package-handler/:config` (https://github.com/jwiegley/use-package/blob/4fa73b1/use-package-core.el#L1471):

    (if (or (null arg) (equal arg '(t)))
        body
      (use-package-with-elapsed-timer
       (format "Configuring package %s" name-symbol)
       (funcall use-package--hush-function :config
                (use-package-concat
                 (use-package-hook-injector
                  (symbol-name name-symbol) :config arg)
                 body
                 (list t)))))

If there is no `:config`, it seems in my case anyway `arg` is indeed `(t)` and `use-package-hook-injector` never runs.  I'm not sure if this is a doc bug or a code bug -- I'm hoping the latter, because as a spacemacs user, that would stop me falling back to `with-eval-after-load` when `spacemacs|use-package-add-hook` "doesn't work" (this is the function spacemacs exposes for users to add these use-package inject hooks).  I'd try to make a PR, but I'm not sure what the intent of that if test above is?

I can reproduce the issue by adding a `:post-config` inject hook using recent spacemacs `develop` branch (https://github.com/syl20bnr/spacemacs/commits/3747afb), which is using `use-package` 20190529.2040.  Here's the use-package declaration:

https://github.com/syl20bnr/spacemacs/blob/3747afb/layers/+source-control/git/packages.el#L49

That gets called after my use of `spacemacs|use-package-add-hook` in my personal config, which successfully adds `use-package--evil-magit--post-config-hook`.  It looks like this:

    (defun jjl/pre-init-evil-magit ()
      (message "### jjl/pre-init-evil-magit")
      (spacemacs|use-package-add-hook evil-magit
        :post-config
        (message "### jjl/pre-init-evil-magit post-config")
        (spacemacs/set-leader-keys
          "gb" 'magit-blame)
        )

The spacemacs use-package declaration for that package has no :config property.  If I add a `:config` that just does `(progn nil)`, my post-config hook does get run (I see the second `message` output above).  That seems to be contrary to the docs.

I suspect this will be confusing a lot of "naive" spacemacs users who (like me until an hour ago) have not figured out what's going on here.

jjlee:
Fixes #785 

I couldn't figure out what the special `(t)` value is for `:config`, so I left that in even though it seems to have identical effect to `nil`.

### DIFF
--- a/use-package-core.el
+++ b/use-package-core.el
@@ -156,8 +156,7 @@ See also `use-package-defaults', which uses this value."
   :group 'use-package)
 
 (defcustom use-package-defaults
-  '(;; this '(t) has special meaning; see `use-package-handler/:config'
-    (:config '(t) t)
+  '((:config nil t)
     (:init nil t)
     (:catch t (lambda (name args)
                 (not use-package-expand-minimally)))
@@ -1468,7 +1467,7 @@ no keyword implies `:all'."
     (use-package-concat
      (when use-package-compute-statistics
        `((use-package-statistics-gather :config ',name nil)))
-     (if (or (null arg) (equal arg '(t)))
+     (if (and (null arg) (not use-package-inject-hooks))
          body
        (use-package-with-elapsed-timer
            (format "Configuring package %s" name-symbol)
--- a/use-package-core.el
+++ b/use-package-core.el
@@ -156,7 +156,8 @@ See also `use-package-defaults', which uses this value."
   :group 'use-package)
 
 (defcustom use-package-defaults
-  '((:config nil t)
+  '(;; this '(t) has special meaning; see `use-package-handler/:config'
+    (:config '(t) t)
     (:init nil t)
     (:catch t (lambda (name args)
                 (not use-package-expand-minimally)))
@@ -1467,7 +1468,7 @@ no keyword implies `:all'."
     (use-package-concat
      (when use-package-compute-statistics
        `((use-package-statistics-gather :config ',name nil)))
-     (if (and (null arg) (not use-package-inject-hooks))
+     (if (and (or (null arg) (equal arg '(t))) (not use-package-inject-hooks))
          body
        (use-package-with-elapsed-timer
            (format "Configuring package %s" name-symbol)
--- a/use-package-tests.el
+++ b/use-package-tests.el
@@ -1474,6 +1474,36 @@
              (config)
              t))))))
 
+(ert-deftest use-package-test/pre-post-hooks-with-:config ()
+  (let ((use-package-inject-hooks t))
+    (match-expansion
+     (use-package foo :config (config))
+     `(progn
+       (when
+           (run-hook-with-args-until-failure 'use-package--foo--pre-init-hook)
+         (run-hooks 'use-package--foo--post-init-hook))
+       (require 'foo nil nil)
+       (when
+           (run-hook-with-args-until-failure 'use-package--foo--pre-config-hook)
+         (config)
+         (run-hooks 'use-package--foo--post-config-hook))
+       t))))
+
+(ert-deftest use-package-test/pre-post-hooks-without-:config ()
+  ;; https://github.com/jwiegley/use-package/issues/785
+  (let ((use-package-inject-hooks t))
+    (match-expansion
+     (use-package foo)
+     `(progn
+        (when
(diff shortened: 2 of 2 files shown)


jjlee:
Note for spacemacs people:
* the tests do pass (use-package Travis is broken right now)
* the tests should be run with `use-package-inject-hooks` set to `nil`

emacs18:
It appears that the fix made on Jun 18 did not last long.
On Jun 29 *b9f1fe6* was checked in to merge *c697e62* which reverted the fix made on Jun 18.
Thus as Jun 29 https://github.com/syl20bnr/spacemacs/issues/13545 resurfaced.
The problem is that following now needs to be moved very early during spacemacs initialization:
```elisp
(custom-set-variables '(use-package-inject-hooks t))
```
If I were to revert *c697e62* on my copy of use-package (but otherwise use the latest version), then the spacemacs problem is resolved.  Thus either *c697e62* needs to be reverted or spacemacs must be changed to set `use-package-inject-hooks` to `t` much sooner.

jjlee:
Hi @emacs18 

When you say "revert", it sounds like you mean that you started seeing the symptoms of your spacemacs bug https://github.com/syl20bnr/spacemacs/issues/13545 again?  "Revert" normally means that a code change itself (rather than its effects) is reversed -- but the commit you reference (https://github.com/jwiegley/use-package/commit/b9f1fe64ee11cf99d84857d51ac74da36a0b744c) does not do that, and in fact I don't see the relevance of that commit to this issue you're commenting on here.

I'm also not sure that use-package is even at fault for your spacemacs issue -- so I suggest talking about that on the spacemacs tracker rather than here.

If you do decide that use-package does need a fix, I guess you should open a new issue for that.

Thanks

emacs18:
I don't know whether this is a use-package problem or spacemacs problem.  Here is what I know
- use-package version between June 18 and June 29 works fine, i.e., the version with your fix but prior to b9f1fe6
- use-package version before June 18 and after June 29 does not work
- use-package version of today minus b9f1fe6 also work.  By this I mean the following
  git checkout master
  git show c697e62 | patch -p1 -R

With the same latest spacemacs, the problem is associated with use-package change c697e62.

jjlee:
`I don't know whether this is a use-package problem or spacemacs problem.`

For that reason, please don't post to old use-package github issues about it where the use-package maintainer might have to pay attention to it: please instead keep it to the spacemacs bug until somebody has verified it's actually a use-package bug.  Thanks

emacs18:
Sorry.   It sure looked like use-package problem, but I am no use-package expert.

skangas:
@jjlee We seem to be missing a [copyright assignment](https://www.gnu.org/licenses/why-assign.html) from you for GNU Emacs. This commit is above the limit of what we could consider exempt from that requirement. Would you be willing to sign the papers, so that we can merge your contribution into Emacs itself? If so, please follow these instructions:

```
Please email the following information to assign@gnu.org, and we
will send you the assignment form for your past and future changes.

Please use your full legal name (in ASCII characters) as the subject
line of the message.
----------------------------------------------------------------------
REQUEST: SEND FORM FOR PAST AND FUTURE CHANGES

[What is the name of the program or package you're contributing to?]
GNU Emacs

[Did you copy any files or text written by someone else in these changes?
Even if that material is free software, we need to know about it.]

[Do you have an employer who might have a basis to claim to own
your changes?  Do you attend a school which might make such a claim?]

[For the copyright registration, what country are you a citizen of?]

[What year were you born?]

[Please write your email address here.]

[Please write your postal address here.]

[Which files have you changed so far, and which new files have you written
so far?]
```

We are finishing up merging `use-package` into Emacs core, so I'm inclined to revert the **test** part of this until that is done. That way we can mark the rest of the commits as "copyright paperwork exempt". @jwiegley Thoughts?

See also issue #282.

PR #846, opened 2020-06-09
Support keymap symbol in bind-key. Fix #845

### LINKED ISSUE #845: describe-personal-bindings error when keymap is dynamically constructed
Reproduction:
1. In a mode hook, do 
```elisp
(bind-key "C-c f" 'ignore (symbol-value (intern (concat (symbol-name major-mode) "-map"))))
```
2. `M-x describe-personal-bindings`
3. `Symbol’s value as variable is void: keymap` in echo area. No personal binding buffer displayed.

Expectation:
No error and the personal binding buffer is displayed.

wyuenho:
Sometimes, there are packages such as [prettier](https://github.com/jscheid/prettier.el) that expose commands that one would like to bind to multiple modes in their mode hooks. In which case the mode map will likely be dynamically constructed from the mode symbol, and one would like to bind keys like the following:

```elisp
(dolist (mode '(css-mode js-mode ...))
  (let ((mode-hook (derived-mode-hook-name mode)))
    (add-hook mode-hook (lambda ()
       (bind-key "C-c f" 'prettier-prettify (derived-mode-map-name mode)))))
```
The problem is,  `bind-key` currently does not support `KEYMAP` supplied as a symbol, and if it is supplied as a `(symbol-value (derived-mode-map-name mode))`, there's no way to reverse lookup the name of the variable from the symbol value later on when constructing the buffer for `describe-personal-bindings`, hence #845 .

This PR proposes a fix to this problem by supporting KEYMAP as symbols, while preserving backward compatibility.

### DIFF
--- a/bind-key.el
+++ b/bind-key.el
@@ -154,10 +154,12 @@ spelled-out keystrokes, e.g., \"C-c C-z\". See documentation of
 
 COMMAND must be an interactive function or lambda form.
 
-KEYMAP, if present, should be a keymap and not a quoted symbol.
+KEYMAP, if present, should be a keymap variable or symbol.
 For example:
 
   (bind-key \"M-h\" #'some-interactive-function my-mode-map)
+
+  (bind-key \"M-h\" #'some-interactive-function 'my-mode-map)
 
 If PREDICATE is non-nil, it is a form evaluated to determine when
 a key should be bound. It must return non-nil in such cases.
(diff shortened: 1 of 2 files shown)


wyuenho:
@jwiegley How does this look?

jwiegley:
If this is backward compatible with how things worked before, then it sounds good to me.

wyuenho:
It is :)

PR #849, opened 2020-06-21
Remove uses of deprecated ‘flet’ macro

phst:


tarsius:
You should also talk about why this allows the removal of `plist-delete`--if that is indeed the case. Maybe that should be done in a separate commit.

phst:
Yeah, I've extended the commit message a bit to talk about this.

Issue #517, opened 2017-11-22
Advantages of :custom?


DamienCassou:
I've just discovered `:custom` and am wondering the advantages of it. A big problem I see is that when using `:custom`, a call to `customize-variable` on any configured variable followed by `"Save for future session"` will add **all** configured variables to the main `custom-set-variables`. This makes all these variables configured in 2 places now.

The only advantage I see is that the new syntax is a bit shorter than using `setq`.

Is there something else?

jwiegley:
It's really just an alternative to people calling `custom-set-variable` in their `:init` block.  It's not for people (like myself) who use the M-x customize interface, which will have the behavior you mentioned.

jwiegley:
Note that one thing we could try to do is to make [initsplit](https://github.com/dabrahams/initsplit) smart enough to put customizations into `use-package` blocks that it finds in `init.el`, rather than into whole files. @dabrahams might have some thoughts on this.

jwiegley:
Here is some experimental support code:

``` elisp
;; (use-package abbrev
;;   :custom-re "\\`abbrev-"     ; this is inferred as such if not specified
;;   :custom
;;   (abbrev-file-name "~/.emacs.d/abbrevs")
;;   :hook
;;   (prog-mode . abbrev-mode))

(defcustom use-package-init-files (list user-init-file)
  "*List of user files containing use-package declarations."
  :type '(repeat (file :must-match t))
  :group 'use-package)

(defun use-package-custom-save-all ()
  (dolist (file use-package-init-files)
    (with-current-buffer (find-file-noselect file t)
      (use-package-replace-custom-variables)
      (save-buffer))))

(defun use-package-to-string (decl &optional newlined-keywords)
  (with-temp-buffer
    (let ((standard-output (current-buffer)))
      (prin1 decl)
      (goto-char (point-min))
      (while (re-search-forward "\\(\\s-+\\):\\(\\S-+\\)\\(\\s-+\\)" nil t)
        (let ((key (match-string 2)))
          (replace-match "\n" nil t nil 1)
          (if (member key newlined-keywords)
              (replace-match "\n" nil t nil 3)))))
    (emacs-lisp-mode)
    (indent-region (point-min) (point-max))
    (buffer-string)))

(defun use-package-replace-custom-variables ()
  (interactive)
  (save-excursion
    (goto-char (point-min))
    (while (re-search-forward
[...]

PR #850, opened 2020-06-22
use-package-core.el: use the Emacs set-default function to avoid saving :custom vars twice

tzz:
This was discussed in #517 and I CC-ed @jwiegley on the emacs-devel discussion.

The change will ensure that `:custom` vars are not also saved to custom.el.

In my testing it worked correctly, I don't see custom.el entries for those variables anymore. Testing is welcome!

### DIFF
--- a/use-package-core.el
+++ b/use-package-core.el
@@ -1394,7 +1394,9 @@ no keyword implies `:all'."
               (comment (nth 2 def)))
           (unless (and comment (stringp comment))
             (setq comment (format "Customized with use-package %s" name)))
-          `(customize-set-variable (quote ,variable) ,value ,comment)))
+          `(funcall (or (get (quote ,variable) 'custom-set) #'set-default)
+                    (quote ,variable)
+                    ,value)))
     args)
    (use-package-process-keywords name rest state)))
 


rprimus:
Tue Jun 30 11:14:51 BST 2020

HI,

`customize-set-variable` is still mentioned in the docstring for the macro `use-package`.  It is still used in `use-package-test/:custom-1`.

```
:custom          Call `customize-set-variable' with each variable definition.
```

link0ff:
@tzz `:custom-face` needs a similar fix where `custom-set-faces` should be replaced with `face-spec-set`.

tzz:
> @tzz This broke the test. In addition, you need to fix the docstring too. Please fix them.

Thank you, and apologies for the late update. Please see #855.

tzz:
> @tzz `:custom-face` needs a similar fix where `custom-set-faces` should be replaced with `face-spec-set`.

@link0ff thank you for the suggestion. I don't know the Emacs face code at all, and looking at `custom-theme-set-faces` I'm concerned it's fairly complex in the interactions with themes.

link0ff:
> looking at `custom-theme-set-faces` I'm concerned it's fairly complex in the interactions with themes.

Actually you don't need all the complexity of `custom-theme-set-faces`, like you don't need all the complexity of `customize-set-variable`.  Exactly like you replaced `customize-set-variable` with just `set-default` that in terms of the Customization UI means that the value is changed outside of the Customization UI, the same way it should be possible to replace `custom-set-faces` with just `face-spec-set` with the meaning that the face is set outside of the Customization UI.

a13:
This change had broken old `:custom` behavior in a lot of ways, see #856 and #861  in particular.

a13:
> The change will ensure that :custom vars are not also saved to custom.el.

why do you use `custom.el` at all, if you already use `:custom` keyword?

akhramov:
Sadly this breaks my workflows. Not to say the change should be rolled back, yet something needs to be figured.

I'll follow up.

a13:
> Those two are not exclusive.

of course, I'm wondering /why/ one may want to use both

link0ff:
Unfortunately, `set-default` is not an adequate replacement of `customize-set-variable` because when a command calls `custom-reevaluate-setting`, it reverts such user setting to the default value, not to the value set by the user with `set-default`.

And there are many such commands in Emacs that use `custom-reevaluate-setting` to revert variables to customized values.

link0ff:
See also https://lists.gnu.org/archive/html/emacs-devel/2020-09/msg00306.html
So custom should be fixed to save package customizations to the same place where these customizations were loaded, i.e. in the `:custom` keyword of `use-package`.

tzz:
@link0ff please see #881 for an attempt to resolve this issue with regards to custom variables. I think a similar fix for custom faces is possible but am not sure, considering this is the filter applied to them:

```
	  (when (or (and spec (eq (nth 0 spec) 'user))
		    comment
		    (and (null spec) (get symbol 'saved-face)))
```

What do you think?

Thanks to everyone who chimed in, and apologies for the late followup.

link0ff:
Thanks, it seems this is the right fix for now (later custom.el could be improved to save customizations to the same place where they were loaded, but this is much larger redesign of custom.el).

Regarding the custom faces: like you put `theme-value` in your patch, perhaps for faces it should work when you put the same way its counterpart symbol `theme-face`.

PR #874, opened 2020-10-13
Fix MELPA links

wasamasa:
melpa.milkbox.net is no more, melpa.org it is now.


wasamasa:
Closing in favor of https://github.com/jwiegley/use-package/pull/873

Issue #877, opened 2020-11-06
describe-personal-keybindings shows `nil` bindings


minad:
Bindings which were unbound using `bind-key "..." nil` or `unbind-key` are still shown in the `describe-personal-keybindings` buffer. Can someone confirm this? If there is interest in a fix, I can work on a PR.

minad:
Closing this in favor of #884, #910.

PR #879, opened 2020-11-07
fix 877: remove nil keybindings from personal-keybindings

minad:
works for both bind-key and unbind-key

See #877


minad:
Closing this in favor of #910

PR #897, opened 2020-12-27
update melpa url

trobol:


trobol:
duplicate of https://github.com/jwiegley/use-package/pull/873

PR #899, opened 2021-01-09
Remove use-package theme from global list of custom-enabled-themes

tzz:
* eval-when-compile for cases where use-package is only required at
  compile time

* remove the 'use-package theme from custom-enabled-themes so e.g.
  (mapc #'disable-theme custom-enabled-themes)
  won't kill user settings.

### DIFF
--- a/use-package-core.el
+++ b/use-package-core.el
@@ -43,10 +43,15 @@
 (require 'cl-lib)
 (require 'tabulated-list)
 
-;; Declare a synthetic theme for :custom variables.
-;; Necessary in order to avoid having those variables saved by custom.el.
-(deftheme use-package)
+(eval-and-compile
+  ;; Declare a synthetic theme for :custom variables.
+  ;; Necessary in order to avoid having those variables saved by custom.el.
+  (deftheme use-package))
+
 (enable-theme 'use-package)
+;; Remove the synthetic use-package theme from the enabled themes, so
+;; iterating over them to "disable all themes" won't disable it.
+(setq custom-enabled-themes (remq 'use-package custom-enabled-themes))
 
 (if (and (eq emacs-major-version 24) (eq emacs-minor-version 3))
     (defsubst hash-table-keys (hash-table)


tzz:
@terlar @lbolla @gusbrs please review. I am not sure about the compile-time evaluation but I think it's correct.

I think this approach (removing `'use-package` from `custom-enabled-themes`) is the best approach to non-invasively resolve the issues you raised, while at the same time avoiding changes and feature requests to other packages and to the Emacs core.

This approach also avoids advising functions, which is always fragile as @gusbrs pointed out.

lbolla:
@tzz I tried the patch and it seems to be working fine for my use case. I am not an expert elisper though, so I am not sure if manually managing custom-enabled-themes may have other consequences.

tzz (commented):
The double `deftheme` call makes the tests pass but I'd love to know if there's a better way.

lbolla (dismissed):
Looking good to me.

kljohann:
> The double `deftheme` call makes the tests pass but I'd love to know if there's a better way.

Maybe `eval-and-compile` does what you want?

tzz:
> Maybe `eval-and-compile` ("and" instead of "when") does what you want?

Perfect; done.

arminfriedl:
Incidentally this would also fix a quite annoying behavior of `customize-themes` with the synthetic theme in place. Usually when only one custom theme is enabled, `customize-themes` will switch between themes, disabling the old one. With the synthetic theme in place however `customize-themes` will overlay themes (because `(> (length custom-enabled-themes) 1)`). 

Here's the relevant code in `cus-theme.el`:
```elisp
(defun customize-themes (&optional buffer)
  [...]
  (make-local-variable 'custom-theme-allow-multiple-selections)
  (and (null custom-theme-allow-multiple-selections)
       (> (length custom-enabled-themes) 1)
       (setq custom-theme-allow-multiple-selections t)
```

When removing `use-package` synthetic theme from `custom-enabled-themes`, `custom-theme-allow-multiple-selections` would work more in line with how was meant to work.

jwiegley (approved):

Issue #812, opened 2019-12-06
`let' around `use-package' not changes hook suffix


andreyorst:
I kinda like, that I don't have to type `-hook` every time in my whole configuration, but some packages use different names for hooks, and I'd like to set those up with different suffix.

This issue is related to #530 and 004ee52fc812938ed42eba8a09e815299ebcead2. Unfortunately I don't understand how to use this customization options as locally as possible. My first thought was to use `let` around `use-package` macro, but this did not worked for me (I've omitted some code):

```emacs-lisp
(let ((use-package-hook-name-suffix "-functions"))
  (use-package vterm
    :bind (("C-`" . aorst/vterm-toggle)
           ("C-t" . aorst/vterm-focus))
    :hook (vterm-exit . aorst/kill-vterm)
    :config
    (defun aorst/vterm-toggle (&optional arg)
      "Toggle `vterm' window on and off with the same command." ...)
      (defun aorst/vterm-focus (&optional arg)
        "Focus `vterm' or open one if there's none." ...)
      (defun aorst/kill-vterm (buf)
        "Kill the `*vterm*' buffer after shell exits."
        (when buf (kill-buffer buf))))))
```

After expansion I have `vterm-exit-hook` variable, not `vterm-exit-functions` variable. Is this a bug, or do I have to configure it in a different manner? If I set suffix to `nil` the default `-hook` suffix is still used in this example.

My second attempt was to use `setq` around the `use-package`:
```emacs-lisp
(setq use-package-hook-name-suffix "-functions")
(use-package vterm
  :bind (("C-`" . aorst/vterm-toggle)
         ("C-t" . aorst/vterm-focus))
  :hook (vterm-exit . aorst/kill-vterm) ...)
(setq use-package-hook-name-suffix "-hook")
```

And this worked, however I don't like this approach (this is why we have `let` in the end) and don't quite understand why `let` doesn't work the same way in this particular example, but seems to work in the test, added in 004ee52fc812938ed42eba8a09e815299ebcead2.

I'm using Emacs 27 built myself from master, use-package melpa version `20191126.2034`, and `lexical-binding` is set to `t`.

Maybe `:hook-suffix` could be added?

```emacs-lisp
(use-package package
  :hook-suffix "special"
  :hook (hook . foo))
```

would produce `hook-special`.

conao3:
use-package is a macro, you should careful with it.
I can generate Sexp that you want to use below snippet.

```elisp
[...]

PR #914, opened 2021-02-22
Add :hook* keyword

conao3:
This keyword is proposed at #812.  Special thanks for @andreyorst.

NOTE:
This PR may not completed.  I don't familier with use-package
codebase, I cannot find code that suppress `require` statemant
when :hook specified.


conao3:
I mistook git operation, this PR HEAD is my master branch. Open another PR to fix this, #916.

PR #828, opened 2020-04-01
Accept vector as key-name in unbind-key

conao3:
This PR resolve issue reported at #827.
Convert the vector to a string in the same way as bind-key and
compare it to the value of `personal-keybindings`

PR #916, opened 2021-02-22
Add :hook* keyword

conao3:
This keyword is proposed at #812. Special thanks for @andreyorst.

NOTE:
This PR may not completed. I don't familier with use-package
codebase, I cannot find code that suppress require statemant
when :hook specified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,6 +358,16 @@ functions being hooked to implicitly be read as `:commands` (meaning they will
 establish interactive `autoload` definitions for that module, if not already
 defined as functions), and so `:defer t` is also implied by `:hook`.
 
+When you want to ignore `use-package-hook-name-suffix`, please
+try `:hook*` keyword like `:hook` keyword.
+
+```elisp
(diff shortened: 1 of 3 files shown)


jwiegley:
I like this PR, just not so much the name. In other places, `*` means global or "in a different context". Here what you really mean is `:hook-raw` or something that does not do any magic when applying the meaning of the keyword.

conao3:
Same as #828, I couldn't take time on this, honestly.
This code does not have to be committed with my name, so if anyone is interested in doing the rest of the work, I would appreciate it.

conao3:
The license notation may be important. I declare this code as CC0 so that you can freely use it.

skangas:
> The license notation may be important. I declare this code as CC0 so that you can freely use it.

Since you have already assigned copyrights to the FSF for your contributions to Emacs, there is no problem for anyone else to re-use this work for that purpose. So the above is not necessary. That said, the gesture is appreciated.

Issue #851, opened 2020-06-28
FR: way to autoload non-interactive functions/macros


aspiers:
I looked for a convenient equivalent to `:commands` which creates the autoloaded function(s) as non-interactive (so that they don't pollute `M-x` completion from `counsel-M-x`, `amx`, etc.) but couldn't find one.  I know I can achieve this manually, e.g.

```
(use-package org-crypt
  :requires org
  :ensure nil
  :defer t
  :init
  (autoload 'org-crypt-use-before-save-magic "org-crypt"))
```

but it would be much nicer if I could write:

```
(use-package org-crypt
  :requires org
  :ensure nil
  :functions org-crypt-use-before-save-magic)
```

aspiers:
Just realised this applies to macros too.

aspiers:
... so maybe the keyword should be `:autoloads` rather than `:functions`.

conao3:
How about?
```elisp
(use-package org-crypt
  :commands org-crypt-use-before-save-magic)
```

It should generate autoload statement that you want

```elisp
(setq use-package-expand-minimally t)
;;=> t

(macroexpand-1
 '(use-package org-crypt
    :commands org-crypt-use-before-save-magic))
;;=> (unless (fboundp 'org-crypt-use-before-save-magic)
;;     (autoload #'org-crypt-use-before-save-magic "org-crypt" nil t))
;;   nil
```

aspiers:
That's what I already tried, and as described above it created an interactive autoload.  The whole point of this issue was to find an alternative which doesn't pollute the list of interactive commands which can be completed after `M-x`.  Having said that, it seems from your macroexpand and from checking my running emacs that maybe this problem was already fixed, because I don't see the issue any more.  Very odd.

conao3:
Oh, sorry I misunderstood your issue. I showed `use-pacakge` generates *interactive* autoload so you do `M-x`, you should see the command (but the function is not interactive)

I think, there is no keyword that you want.
Maybe PR to org-crypt to add `;;;###autoload` comment if you want to autoload...?

aspiers:
[**@conao3**](https://github.com/conao3) commented on [February 22, 2021 4:04 PM](https://github.com/jwiegley/use-package/issues/851#issuecomment-783481749 "2021-02-22T16:04:43Z - Replied by Github Reply Comments"):
[...]

PR #917, opened 2021-02-22
Add :autoload keyword

conao3:
Add `:command*` keyword proposed at #851.  Special thanks to @aspiers.

This `:command*` is similar to `:command`, but it generate autoload
statement as no-interactive one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,13 @@ and within the `isearch-mode-map` (see next section).  When the package is
 actually loaded (by using one of these commands), `moccur-edit` is also
 loaded, to allow editing of the `moccur` buffer.
 
+If you autoload no-interactive function, please use `:autoload`.
+
+```elisp
+(use-package org-crypt
+  :autoload org-crypt-use-before-save-magic)
+```
+
 ## Key-binding
 
 Another common thing to do when loading a module is to bind a key to primary
--- a/use-package-core.el
+++ b/use-package-core.el
@@ -94,6 +94,7 @@
     ;; Any other keyword that also declares commands to be autoloaded (such as
     ;; :bind) must appear before this keyword.
     :commands
+    :autoload
     :init
     :defer
     :demand
@@ -118,7 +119,8 @@ declaration is incorrect."
 (defcustom use-package-deferring-keywords
   '(:bind-keymap
     :bind-keymap*
-    :commands)
+    :commands
+    :autoload)
   "Unless `:demand' is used, keywords in this list imply deferred loading.
 The reason keywords like `:hook' are not in this list is that
 they only imply deferred loading if they reference actual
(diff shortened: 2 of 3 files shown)


thomasf:
There should be a less confusing name than `:command*` for this. IIRC, commands in Emacs terminology must always be interactive.

thomasf:
Function: commandp object &optional for-call-interactively
This function returns t if object is a command. Otherwise, it returns nil.

Commands include strings and vectors (which are treated as keyboard macros), lambda expressions that contain a top-level interactive form (see Using Interactive), byte-code function objects made from such lambda expressions, autoload objects that are declared as interactive (non-nil fourth argument to autoload), and some primitive functions. Also, a symbol is considered a command if it has a non-nil interactive-form property, or if its function definition satisfies commandp.

If for-call-interactively is non-nil, then commandp returns t only for objects that call-interactively could call—thus, not for keyboard macros.

conao3:
Completely agree. I renamed `:commands*` to `:autoload`.

aspiers:
[**@conao3**](https://github.com/conao3) commented on [February 22, 2021 7:13 PM](https://github.com/jwiegley/use-package/pull/917#issuecomment-783607201 "2021-02-22T19:13:40Z - Replied by Github Reply Comments"):
> Completely agree. I renamed `:commands*` to `:autoload`.

As suggested in #851 I think `:functions` is a better choice of symbol here.  `:autoload` is a confusing choice to me because `:commands` already causes autoloads, so there is an undesirable overlap between the two if this name is used.

conao3:
Thanks, @aspiers. Now I think `:functions` is better. How about @thomasf?

CC: @jwiegley

thomasf:
I don't think that is a clean orthogonal feature regardless of what we name it. `:autoload` is at least somewhat self explanatory even if `:commands` also creates autoloads (for commands).  I'm not sure if I think that one argument is particularly stronger than the other here.

aspiers:
[**@thomasf**](https://github.com/thomasf) commented on [February 22, 2021 7:21 PM](https://github.com/jwiegley/use-package/pull/917#issuecomment-783612465 "2021-02-22T19:21:48Z - Replied by Github Reply Comments"):
> I don't think that is a clean orthogonal feature regardl ess of what we name it. `:autoload` is at least somewhat self explanatory even if `:commands` also creates autoloads (for commands.   I'm not sure if I think that one argument is particularly stronger than the other here.

You are right, my suggestion isn't much better :laughing:  Hmmm, I wonder if there is a better word ...

jwiegley:
`:functions` is already use to satisfy the byte-compiler, but we could easily make that keyword do this as well.

jwiegley (approved):

jwiegley:
Thank you, I think this is a good and safe addition.

PR #930, opened 2021-04-17
feat(use-package-ensure-system-package): add update-custom-packages command

danielpza:
Add `use-package-ensure-system-package-update-custom-packages` for updating custom packages not installed by system package manager. It will run the same commands used for installing them.


chuxubank:
I'm curious about should we call `use-package-ensure-system-package-update-custom-packages` by ourself? 
I do not see it being called in the other part of the use-package codes.

I see, it is `(interactive)`, sorry.

PR #968, opened 2021-12-21
Cleanup manual and regenerate use-package.texi

tarsius:
I originally added the manual and just had another look.  There was a lot of cruft laying around and I removed that.

The texi file also had not been regenerated in a while.

Due to the cleanup it should now be possible for everyone to update the texi file by running `make texi` or if the intention is to amend to the current head commit, then `AMEND=t make texi`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,12 @@
-/use-package.info
 *~
-*.elc
+/*-autoloads.el
+/*.elc
+/*.html
+/*.info
+/*.pdf
+/*.texi
+/config.mk
 /dir
 /doc/content/*
+/stats/
+/use-package/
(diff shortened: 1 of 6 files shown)


jwiegley:
Hi @tarsius, the `use-package.texi` was just recently updated. Should this PR overwrite?

tarsius:
I've rebase this pr and fixed it up a bit. It can be merged without loss of prior work (well I removed some stubs that didn't get completed in years, but that's it).

By the way, it is now possible to regenerate the manual with stock Org.

If you are interested, I could also trim down the makefiles. They contain a lot of Magit-specific stuff (which is where you copied it from). I've already started working on that.

jwiegley:
@tarsius That cleanup would be great, thank you!

jwiegley:
@tarsius That cleanup would be great, thank you!

Issue #991, opened 2022-08-08
:bind* not working anymore?


Trisk3lion:
Hello, thanks for a great package!

":bind*" stopped working for me after updating to the last version of bind-key. So something in the latest commits probably did it.

I haven't looked into it more then that yet thou..

cxa:
Downgrade package `bind-key` or wait for new update.

(P.S. the simple way is copy source from `Download` button on <https://stable.melpa.org/#/bind-key>, override `~/.emacs.d/elpa/bind-key-20220807.1553/bind-key.el`，and delete `bind-key.elc` in that dir.

pegzmasta:
I hope they update it soon.  I have a lot of ":bind*" tags in my startup file, and they no longer work.

realcomplex:
For what it's worth, I submitted a one-line pull-request (#993) which fixes the issue. You could try to use my fork for now or just apply the fix yourself. Or really, the problem is that `override-global-mode` isn't turned on automatically anymore like before, so turning that minor mode on manually in your config should also do the trick.

jwiegley:
I just noticed this as well, pulling in the fix now.

jwiegley:
Thank you for addressing this so quickly @realcomplex.

PR #993, opened 2022-08-10
Make sure that bind-key's `override-global-mode` is initially on

realcomplex:
In cb90d3f the arguments to `define-minor-mode` were changed
erroneously. Whereas the `override-global-mode` was initially defined
as `(define-minor-mode override-global-mode "..." t "")`, the two
latter arguments where changed to `:global t :lighter ""`. However,
the two original arguments corresponded to the keywords `:init-value`
and `:lighter`, respectively.

With `:init-value t` missing, the minor mode isn't enabled by default,
and `bind-key*` appears not to work (see also #991). This fixes the
issue.

Note that to follow the exact meaning of the original code, one would
also have to remove the `:global t` argument. I don't know if that's
desirable or not, probably not.

### DIFF
--- a/bind-key.el
+++ b/bind-key.el
@@ -132,7 +132,6 @@
 (define-minor-mode override-global-mode
   "A minor mode so that keymap settings override other modes."
   :init-value t
-  :global t
   :lighter "")
 
 ;; the keymaps in `emulation-mode-map-alists' take precedence over
--- a/bind-key.el
+++ b/bind-key.el
@@ -131,6 +131,7 @@
 
 (define-minor-mode override-global-mode
   "A minor mode so that keymap settings override other modes."
+  :init-value t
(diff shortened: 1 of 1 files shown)


piyo:
Why is :global t being set? This is a code refactor so please remove it from your PR.

realcomplex:
Yes that makes sense, like I said I didn't know if there was actually any intention behind making the minor mode global in cb90d3f but let's stick to the original meaning of the code.

chadhs:
Just a heads up, my global overrides stopped working as of this update.  I had to explicitly `(setq override-global-mode t)` to make them work again.  I can provide more info if necessary.

jwiegley:
That's curious, so the `:init-value t` setting does not set the initial value to `t`?

PR #1004, opened 2022-10-15
fix #934: use face-spec-set instead of custom-set-faces

### LINKED ISSUE #934: custom-face, custom-file, and org-babel-load-file
I guess that because predicate `custom-face` calls function `customize-set-faces`, the configured values are stored by emacs in the file defined by variable `custom-file`, so the user ends up with the customization present at two separate points in the Emacs init process. This can get confusing because making a change in the future at one point will not in all cases be sufficient, and a user would need to know / remember about the existence of both `custom-file` and the specific `use-package` invocaion.

In my personal config, I've replaced all uses of predicate `custom-face` with calls to `face-spec-set` within the `config` predicate, and removed the parallel entries in my `config-file`.

Maybe `use-package` should call `face-spec-set` instead of `customize-set-faces`?

andreyorst:
For reference, each time I press the "apply and save" button when I use Custom, my custom file stores not only the settings I've made but each face specified with the `:custom-face` section in my `use-package` declarations. This PR fixes it, as suggested in #934.

### DIFF
--- a/use-package-core.el
+++ b/use-package-core.el
@@ -1518,7 +1518,7 @@ no keyword implies `:all'."
 (defun use-package-handler/:custom-face (name _keyword args rest state)
   "Generate use-package custom-face keyword code."
   (use-package-concat
-   (mapcar #'(lambda (def) `(custom-set-faces (backquote ,def))) args)
+   (mapcar #'(lambda (def) `(apply #'face-spec-set (backquote ,def))) args)
    (use-package-process-keywords name rest state)))
 
 ;;;; :init
--- a/use-package-tests.el
+++ b/use-package-tests.el
@@ -1156,7 +1156,7 @@
   (match-expansion
    (use-package foo :custom-face (foo ((t (:background "#e4edfc")))))
    `(progn
-      (custom-set-faces (backquote (foo ((t (:background "#e4edfc"))))))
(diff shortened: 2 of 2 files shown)


jwiegley:
I assume you've applied this change locally and it does the right thing when customize is used?

andreyorst:
> I assume you've applied this change locally and it does the right thing when customize is used?

yes, faces are set properly for me on emacs 29 at least

andreyorst:
@jwiegley I've fixed macro expansion tests that were affected by this change, so I guess everything should be fine now.

jwiegley:
Thank you, I just need to try this under my Emacs 28 next week to see how it does. What about Emacs 26 and 27? Do we need a version check, or will this change work on all of them?

andreyorst:
>What about Emacs 26 and 27? Do we need a version check, or will this change work on all of them?

face-spec-set was introduced in Emacs 20.4 so I guess there should be no problem. I can build Emacs 27 and 26 this weekend and see if everything is ok.



-- 
Andrey Listopadov

jwiegley:
If it was introduced then, don’t worry about it.

jwiegley:
I'm traveling for the next week, but once I'm back and can confirm this locally, I will merge.

PR #787, opened 2019-09-14
':local' keyword adding code to mode-hook

matzebond:
I noticed that I often add a special setup function or lambda to the mode hooks in use-package. Like in the example config of evil-org.
 ```
  (use-package evil-org
    :after (evil org)
    :hook (org-mode-hook . evil-org-mode)
    :config
    (add-hook 'evil-org-mode-hook
	      (lambda ()
	        (evil-org-set-key-theme)))
    (require 'evil-org-agenda)
    (evil-org-agenda-set-keys))
```
I wanted to get rid of this `add-hook` line and came up with the `:local` keyword
 ```
  (use-package evil-org
    :after (evil org)
    :hook (org-mode-hook . evil-org-mode)
    :config
    (require 'evil-org-agenda)
    (evil-org-agenda-set-keys)
    :local
    (evil-org-set-key-theme)
```
In this case it is guessed that the hook is `evil-org-mode-hook`.
This keyword also supports the definition of multiple hooks e.g.
```
  (use-package cc-mode
    :config
    (setq-default c-basic-offset 8)
    :local
    (c-mode-common-hook . ((setq-local tab-width 8)))
    (java-mode-hook . ((setq-local tab-width 4)
                       (setq-local c-basic-offset 4))))
```
Internally this creates `(defun use-package-func/java-mode-hook () ...)` and `(add-hook 'java-mode-hook 'use-package-func/java-mode-hook)` adds this function the correct hook (same for the c-mode-common-hook).

This is my first real elisp coding experience so feel free to point out any non idiomatic or stupid code :)
If you want to merge this PR I can add some documentation in the readme.

Cheers

PR #1027, opened 2022-11-27
Revert "Add: 'local' keyword" for lack of copyright assignment

skangas:
We are still lacking a copyright assignment for @matzebond, which is blocking the merge of `use-package` into Emacs. The only way forward if we want `use-package` in Emacs 29 is to revert this commit for now, and re-merge it once the copyright assignment is done. See issue #787 and #282.

@jwiegley What do you think?

This reverts commit 2808c8466fcce62cc96ae159d96d8205e78c4053.

### DIFF
--- a/use-package-core.el
+++ b/use-package-core.el
@@ -100,8 +100,7 @@
     :load
     ;; This must occur almost last; the only forms which should appear after
     ;; are those that must happen directly after the config forms.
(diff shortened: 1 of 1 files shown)


jwiegley:
We'll revert for now, and seek assignment to reapply.

PR #1019, opened 2022-11-15
Allow multiple keymaps in :map argument

jwiegley:
Originally created as #830.

PR #1051, opened 2023-05-10
Allow multiple keymaps in :map argument

fishyfriend:
This is a reimplementation of #830, released as #1019 but reverted due to lacking copyright assignment. The copyright issue should now be solved as I have signed the Emacs copyright papers.

Sorry for the trouble of requesting a second review. I felt a rewrite was warranted due to discovered loose ends as well as readability.

Noteworthy changes:

- `:map nil`, which currently causes an error, is now treated as synonymous with `:map global-map`.

- `nil`, `global-map`, and `override-global-map` are explicitly disallowed as arguments for `:prefix-map` or `:repeat-map`. Existing checks [here](https://github.com/jwiegley/use-package/compare/master...fishyfriend:maps3?expand=1#diff-4bf2204d9f3a58d301a6b4c1c4efb86a004a3af832ba69eac91d72d94090f506L314) and [here](https://github.com/jwiegley/use-package/compare/master...fishyfriend:maps3?expand=1#diff-4bf2204d9f3a58d301a6b4c1c4efb86a004a3af832ba69eac91d72d94090f506L320) may have been intended for this purpose (perhaps mistakenly substituting `map` for `(cadr args)`?). These checks had simply been removed in the original PR as I couldn't make sense of them.

- A redundant [check](https://github.com/jwiegley/use-package/compare/master...fishyfriend:maps3?expand=1#diff-4bf2204d9f3a58d301a6b4c1c4efb86a004a3af832ba69eac91d72d94090f506L396) related to repeat maps has been removed

## Original description from #830

This PR augments bind-keys to support passing a list of keymaps as the `:map` argument.

When one wants to bind the same key/command pair in multiple keymaps, the current options have some drawbacks:

```elisp
;; Option 1: Code duplication, problematic when duplicated keybindings are numerous
(use-package foopkg
   :bind (:map bar-mode-map
          ("C-c x" . foocmd1)
          ("C-c y" . foocmd2)
          :map baz-mode-map
          ("C-c x" . foocmd1)
          ("C-c y" . foocmd2)))

;; Option 2: Use a loop, sacrificing the convenience and readability of :bind
(use-package foopkg
  :commands (foocmd1 foocmd2)
  :init
  (dolist (map (list bar-mode-map baz-mode-map))
    (bind-keys :map map
               ("C-c x" . foocmd1)
               ("C-c y" . foocmd2))))
```

With the new way, it is possible to avoid duplication and still use `:bind`:

```elisp
(use-package foopkg
  :bind (:map (bar-mode-map baz-mode-map)
        ("C-c x" . foocmd1)
        ("C-c y" . foocmd2)))
```

Additionally this PR clarifies the documentation of `bind-chords` to reflect that this style is already available for that command.

-----

Thanks,
Jacob

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,14 @@ The effect of this statement is to wait until `helm` has loaded, and then to
 bind the key `C-c h` to `helm-execute-persistent-action` within Helm's local
 keymap, `helm-command-map`.
 
+Multiple keymaps can be specified as a list:
+
+``` elisp
+(use-package helm
+  :bind (:map (lisp-mode-map emacs-lisp-mode-map)
+         ("C-c x" . eval-print-last-sexp)))
+```
+
 Multiple uses of `:map` may be specified. Any binding occurring before the
 first use of `:map` are applied to the global keymap:
 
--- a/bind-chord.el
+++ b/bind-chord.el
@@ -102,7 +102,8 @@ function symbol (unquoted)."
   "Bind multiple chords at once.
 
 Accepts keyword argument:
-:map - a keymap into which the keybindings should be added
+:map - a keymap or list of keymaps into which the keybindings should be
+       added
 
 The rest of the arguments are conses of keybinding string and a
 function symbol (unquoted)."
--- a/bind-key.el
+++ b/bind-key.el
@@ -262,12 +262,13 @@ In contrast to `define-key', this function removes the binding from the keymap."
   "Similar to `bind-key', but overrides any mode-specific bindings."
   `(bind-key ,key-name ,command override-global-map ,predicate))
 
-(defun bind-keys-form (args keymap)
+(defun bind-keys-form (args keymaps)
   "Bind multiple keys at once.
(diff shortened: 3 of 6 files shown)


benthamite:
@jwiegley Would it be possible to merge this pull request? It [seems like you were ready to merge the previous version](https://github.com/jwiegley/use-package/pull/1019#issuecomment-1315646218) and the only obstacle was the copyright assignment legal requirement, which is now addressed. Thanks.

fishyfriend:
Hi, original submitter here, responding to @benthamite's comment. IMO a fresh review is needed before merging. Although this PR is not a huge set of changes, the part of bind-key.el involved is pretty complex, the functional changes are not totally identical to the original PR (see description), and the revised PR touches code that moved after the original version was reverted.

benthamite:
@fishyfriend, oh, I see. Thanks for clarifying, and please disregard my previous comment.

PR #1088, opened 2025-03-28
Bind chord use key chord define

LemonBreezes:
Update bind-chord to use the `key-chord-define` instead of hacking the internal representation of a key-chord. This is because I am making deep changes to key-chord in https://github.com/emacsorphanage/key-chord/pull/12.


LemonBreezes:
Closing this in favor of https://github.com/waymondo/use-package-chords/pull/14.

